PR #84, opened 2025-04-03
[nexus-tool] social tool batch #2 (4.1.1)-> POST /2/tweets/:id/retweets - Retweet a tweet

afarukcali:
## Checklist

- [x] write tests

### DIFF
--- a/tools/social-twitter/src/tweet/retweet_tweet.rs
+++ b/tools/social-twitter/src/tweet/retweet_tweet.rs
@@ -6,6 +6,7 @@ use {
     super::models::RetweetResponse,
     crate::{
         auth::TwitterAuth,
+        error::TwitterErrorKind,
         twitter_client::{TwitterClient, TWITTER_API_BASE},
     },
     nexus_sdk::{fqn, ToolFqn},
@@ -37,8 +38,13 @@ pub(crate) enum Output {
         retweeted: bool,
     },
     Err {
-        /// Error message
+        /// Detailed error message
         reason: String,
+        /// Type of error (network, server, auth, etc.)
+        kind: TwitterErrorKind,
+        /// HTTP status code if available
+        #[serde(skip_serializing_if = "Option::is_none")]
+        status_code: Option<u16>,
     },
 }
 
@@ -78,6 +84,8 @@ impl NexusTool for RetweetTweet {
             Err(e) => {
                 return Output::Err {
                     reason: e.to_string(),
+                    kind: TwitterErrorKind::Network,
+                    status_code: None,
                 }
             }
         };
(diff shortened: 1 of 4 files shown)


tools/social-twitter/src/twitter_client.rs
```diff
@@ -0,0 +1,175 @@
...
+        error::{parse_twitter_response, TwitterError, TwitterResult},
+    },
+    reqwest::Client,
+    serde::{de::DeserializeOwned, Serialize},
+    serde_json::Value,
+    std::sync::Arc,
+};
+
+/// Twitter API client for making authenticated requests
+pub struct TwitterClient {
```
oguzhaniptes, line 17: @aelesbao We did the suggestion you mentioned in the last comment. We extracted everything from `twitter_client` module. If this structure is suitable, we will integrate this structure in other branches.

tools/social-twitter/src/tweet/retweet_tweet.rs
```diff
@@ -0,0 +1,382 @@
...
+
+    fn path() -> &'static str {
+        "/retweet-tweet"
+    }
+
+    async fn health(&self) -> AnyResult<StatusCode> {
+        Ok(StatusCode::OK)
+    }
+
+    async fn invoke(&self, request: Self::Input) -> Self::Output {
```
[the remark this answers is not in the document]
oguzhaniptes: https://github.com/Talus-Network/nexus-sdk/pull/84/files#r2050639042

aelesbao:
@oguzhaniptes @afarukcali, could you please update the current branch with the latest version on `main`?

oguzhaniptes:
Current branch updated to latest version of `main` branch.

tools/social-twitter/src/tweet/retweet_tweet.rs
```diff
@@ -0,0 +1,282 @@
...
+                    };
+                }
+            }
+            Err(e) => {
+                let error_response = e.to_string();
+                Output::Err {
+                    reason: error_response,
+                }
+            }
+        }
```
aelesbao, line 113: The pattern I see repeating in all Twitter tools is:

1. Starts with `let response = request(...)`
2. It checks if the request went through: `match response { Ok(http_response), Err(e) }`. The error here are top-level network issues, etc (let's call it level 1).
3. It checks if the response data is valid `match http_response.data { Ok(data), Err(errors) }`. The error here is about invalid HTTP status codes (level 2 error).

This two-level pattern matching to handle errors is generic enough and could be moved to the `TwitterClient` as well.

aelesbao (changes requested):

tools/social-twitter/src/tweet/retweet_tweet.rs
oguzhaniptes: @aelesbao I moved the error-handling structure you described into the `twitter_client` and made it generic. I've applied it to the `retweet_tweet` endpoint as an initial example. If this approach looks good, we can proceed with applying the same pattern to the other endpoints as well.

aelesbao (approved):
Great work!

kouks (approved):

PR #88, opened 2025-04-03
[nexus-tool] social tool batch #2 (4.1.4)-> DELETE /2/users/:id/likes/:tweet_id - Unlike a tweet

oguzhaniptes:
## Notable changes

- Introduced `unlike-tweet` tool to un-like tweets using Twitter API. Updated README.md to include documentation for the new tool, detailing input parameters and output variants. Modified main.rs and mod.rs to integrate the new tool into the existing framework.

## Checklist

- [x] write tests


tools/social-twitter/src/tweet/models.rs
```diff
@@ -655,4 +655,23 @@ pub struct RetweetData {
...
+    /// Errors returned when the request fails
+    #[serde(default)]
+    pub errors: Option<Vec<TwitterApiError>>,
+}
+
+/// Data structure for a successful unlike response
+#[derive(Debug, Deserialize)]
+pub struct UnlikeData {
+    /// Whether the tweet was successfully unliked
+    pub liked: bool,
```
davidrotari19, line 673: This may be confusing. Could you maybe change to some variants

tools/social-twitter/README.md
```diff
@@ -349,6 +349,56 @@ The like operation failed.
...
+
+The ID of the tweet to unlike.
+
+## Output Variants & Ports
+
+**`ok`**
+
+The tweet was successfully unliked.
+
+- **`ok.liked`: [`bool`]** - Confirmation that the tweet was unliked (false)
```
davidrotari19, line 384: This is confusing. Enum with some variants would be better

aelesbao (approved):

tools/social-twitter/README.md
kouks: I think this is always `true`. There is some slightly "weird" behavior of the workflow where at least 1 output port is required. We could consider adding a "unit" output port in the future to avoid this.

But for now I think it's fine.

kouks: What is confusing however, and should be changed is the name of the port - should be `unliked` imo.

afarukcali: https://github.com/Talus-Network/nexus-sdk/pull/88/commits/6eac756622d86b14b6210596257decb8804f301b fixed

kouks (approved):

PR #128, opened 2025-04-14
feat: no input ports in entry groups

kouks:
## Notable changes

- removing redundant mentions of input ports from entry groups (all input ports _must_ be provided anyway)
- renamed `vertex.input_ports` to `vertex.entry_ports` as that's what they are - if specified at all, they're entry

## References

- finally closes https://github.com/Talus-Network/nexus-next/issues/175
- scaffold for https://github.com/Talus-Network/nexus-next/issues/270

## Checklist

- [x] keep a changelog
- [x] write tests


tmsdkeys:
So if I understand how it works, I think it could be good to add one additional term, "source vertex" that differentiates "entry vertex"

- **entry vertex:** a vertex part of an entry group. ONLY relevant in relation to an entry group
- **source vertex**: a vertex that initiates a DAG walk (possibly concurrent) ALWAYS relevant based on DAG definition

as well, as some related terms:
- **entry port**: an input port on an entry vertex that is NEITHER an input port with a default value NOR with an edge leading into it
- ~**entry point**: synonym to entry vertex or source vertex (?)~ (I would stay clear of this term)
- **entry configuration**: (a subset of) source vertices that can initiate one, or many concurrent, DAG walk(s). ALL entry configurations directly define the related entry group

Then an entry group can be defined as follows ...

An "entry group" should consist of ALL vertices in a given DAG entry configuration that: 
a) are _source vertices_
b) can be reached as part of the DAG execution AND have _at least one entry port_, 

Note: the default entry group will thus consist of:
- all source vertices
- all vertices with entry ports

wdyt?

kouks:
Honeslty I think this distinction is not that useful. If _I_ understand it correctly, the only thing it brings is a slightly smarter default entry group inferrence (we can infer that source vertices with no ports should be part of default entry group). But I think this is more confusing that having people specify that in an entry group manually. My thinking is as follows:

- **entry group:** defines a set of vertices that the DAG can begin an execution with (with up to N concurrent walks where N is the number of vertices in the group)
- **entry port:** list of input ports that have no default values nor incoming edges and **must** be provided data if the port's vertex is part of an invoked entry group
- **vertex:** no distinction here - any vertex can be part of an entry group. If a vertex is part of an entry group, has incoming edges and no entry ports, this is a no-op (meaning that it's the same as the vertex not being part of the entry group). All other cases do have effect on the DAG execution (downstream vertex _with_ entry ports or vertex that has no incoming edges will be triggered if part of an entry group).

tmsdkeys:
@kouks So I've ran this conversation through ChatGPT (as a representative of more objective opinion) and it got confused as well, due to the fact that "entry" is conflating two orthogonal concepts: 1) initiating DAG walk and 2) providing configuration further downstream.

It thought that if a vertex that has incoming edges but is still added to the entry group, implied that this vertex would be the source of a DAG walk and that the runtime would ignore the incoming edge(s) in this case in favor of the entry port provided.

I think this is a reasonable assumption from the term "entry group" and thus confusing.

Curious about other's thoughts here as well, cc @Talus-Network/engineering @LKY-stephen

kouks:
I'm probably biased here because the code doesn't implement this distinction directly (between downstream and non-downstream entry ports). Let's hear some opinions from others and then we can adjust the docs.

davidrotari19:
In my opinion the problem arises because the definition is incomplete. You cant say that its the list of all vertices from which execution starts because some of the input in entry group will be immediately consumed(in case its downstream). imho entry group as the set all all the input ports that have to be provided and from which execution is the most suitable. It does not matter that the input port is consumed immediately in downstream tasks, the port still acted as the input

kouks:
> In my opinion the problem arises because the definition is incomplete. You cant say that its the list of all vertices from which execution starts because some of the input in entry group will be immediately consumed(in case its downstream). imho entry group as the set all all the input ports that have to be provided and from which execution starts is the most suitable(in case of vertices with no input ports implicit port is assumed). It does not matter that the input port is consumed immediately in downstream tasks, the port still acted as the input

We've moved away from specifying input ports in entry groups with this PR. It's just vertices now. It is automatically inferred that **all** entry ports that are assigned to vertices that belong to an entry group **must** be provided data for.

davidrotari19:
> We've moved away from specifying input ports in entry groups with this PR. It's just vertices now. It is automatically inferred that **all** entry ports that are assigned to vertices that belong to an entry group **must** be provided data for.

I know that. Maybe I am really missing here something, but what I meant to say is that maybe specific input ports was not a bad ideal. It clearly does not have the same distinction in downstream and non-downstream entry ports, they will all be entry ports. Fundamentally they will start at the same time, even though some will be consumed immediately.

loverdos:
~~Devil's~~ User's advocate. 

What details should the user care about? Are we introducing too much terminology? As a user/caller I just want to be able to say: here is the (input) data, some of them to use immediately (kick it off), some of them to be used down the line for a subset of the Tools but I really need to specify them now because they are fixed values, they are not produced outputs.

davidrotari19:
> > I know that. Maybe I am really missing here something, but what I meant to say is that maybe specific input ports was not a bad idea. It clearly does not have the same distinction in downstream and non-downstream entry ports, they will all be entry ports. Fundamentally they will start at the same time, even though some will be consumed immediately.
> 
> There was no distinction before, either. Entry ports could be downstream or not-downstream. The only difference (complication) was that you had to specify vertex names if the vertex had no entry ports. Which was simplified here.
> 
> Conceptually, this change doesn't _acutally change_ much, it just simplifies the JSON DAG, removing redundant data.

I am fairly sure, we talk about completely different things. At least I did have nothing in mind about change in code, I meant change in definition and naming, so it does not create confusion

LKY-stephen:
> ~Devil's~ User's advocate.
> 
> What details should the user care about? Are we introducing too much terminology? As a user/caller I just want to be able to say: here is the (input) data, some of them to use immediately (kick it off), some of them to be used down the line for a subset of the Tools but I really need to specify them now because they are fixed values, they are not produced outputs.

Agree. The purpose of Nexus is to simplify the workflow consumption. The most straightforward way is to provide only one concept that the user needs to prepare, and let Nexus do all the following tasks. I think it is a bad idea to let people handle the concept inside DAG (vertices and edges) unless they need to debug.

Therefore, I think (and wrote in the whitepaper) there is only one start vertex (can imagine as a default initial onchain tool), and the user activates it to trigger the execution. This start vertex will check all entry ports (as inputs) and trigger all entry vertices(as outputs), and drive the execution. If we can make this start vertex clearly for user (e.g., as a DAG method or Vertex), I think all entry XXX stuff are internal details that most user don't need to worry about.

tmsdkeys:
@LKY-stephen @loverdos 

I think we need to specify "user" here. Is it an agent/workflow consumer or an agent developer (or even an end user)? I agree we should allow for things to be simple and abstracted for agent consumers but at the same time allow for agent developers who want MORE fine grained control to set up complex workflows and entry groups, to do so should they wish... the terminology could be opaque for the former but should be clear for the latter

A good framework -> gives you good defaults but also allows you to go in depth

LKY-stephen:
> @LKY-stephen @loverdos
> 
> I think we need to specify "user" here. Is it an agent/workflow consumer or an agent developer (or even an end user)? I agree we should allow for things to be simple and abstracted for agent consumers but at the same time allow for agent developers who want MORE fine grained control to set up complex workflows and entry groups, to do so should they wish... the terminology could be opaque for the former but should be clear for the latter.
> 
> A good framework -> gives you good defaults behind abstractions but also allows you to peek behind the curtains and exert more control
1. I treat the user as a Talus agent end-user, who doesn't care how the workflow works. 
2. I don't mean there is anything to hide; I just say the external entry should be simple and straightforward, so the user can avoid handling topology.

tmsdkeys:
> 1. I treat the user as a Talus agent end-user, who doesn't care how the workflow works.
> 2. I don't mean there is anything to hide; I just say the external entry should be simple and straightforward, so the user can avoid handling topology.

Okay, but I could have a situation where an agent developer builds a custom TAP with workflows with multiple entry points attached to it with custom business logic to handle potentially 'exotic' entry groups depending on the logic in the contract (package)...

This would not require the end user to be aware of anything but would require the custom TAP agent dev to be able to set up complex flows, is what I mean. We should not determine how best to serve _their_ end users, right?

LKY-stephen:
> ation where an agent developer builds a custom TAP with workflows with multiple entry points attached to it with custom business logic to handle potentially 'exotic' entry groups depending on the logic in the contract (package)...
> 
> This would not require the end user to be aware of anything but would require the custom TAP agent dev to be able to set up complex flows, is what I mean. We should not determine how best to serve _their_ end users, right?

What you described can be a pre-process for the workflow execution. User will have to use TAP method for triggering the talus agent workflow, so the dev can add business logic before that.  Or they can just register the logic as an onchain tool. I don't see there is anything missing here?

tmsdkeys:
> What you described can be a pre-process for the workflow execution. User will have to use TAP method for triggering the talus agent workflow, so the dev can add business logic before that. Or they can just register the logic as an onchain tool. I don't see there is anything missing here?

The main argument that I am countering is: "the Talus agent end user doesn't care about the impl details of DAG and entry groups so this whole discussion doesn't really matter much"

my retort: the end user is already abstracted from this through the layer of the TAP, **so imo the agent developer is the relevant "user" here**. and for them, this is potentially something they want control of.

Example: For a tutorial I've been considering of creating a voting contract (simplified version of DAO gov) where user could vote for a specific workflow to satisfy a task. The different approaches (using different tools or different control flow) are similar but different. Instead of using a new DAG definition for all of the options, I could instead design a composite DAG that simply differentiates multiple entry groups. -> This is a concern for the agent dev, not for the end user (i.e. voter)

My point is that if we want to encourage this agent development design space, entry groups and entry ports need to be understood (and thus, have an intuitive definition). Even if the configuration is indeed processed prior to workflow execution, it's about the design of DAGs with entry groups

LKY-stephen:
> My point is that if we want to encourage this agent development design space, entry groups and entry ports need to be understood (and thus, have an intuitive definition). Even if the configuration is indeed processed prior to workflow execution, it's about the design of DAGs with entry groups

No action/proposal makes this discussion difficult to understand. I am just thinking a native simplified single entry point for agent end user (start vertex or anything you can make it decoupled from the topology complexity). The single entry point can be configured by the agent dev with pre-process(keep topology) or create an onchain tool for modifying the workflow(adjust topology).

.github/workflows/cli.yml
```diff
@@ -13,6 +13,13 @@ on:
       - "cli/**"
       - ".github/workflows/cli.yml"
 
+# Fix for OOM.
+env:
+  CARGO_BUILD_JOBS: 1
+  CARGO_INCREMENTAL: 0
+  RUSTFLAGS: >
+    -C codegen-units=1
```
aelesbao, line 21: From the [rustc book](https://doc.rust-lang.org/rustc/codegen-options/index.html#codegen-units):

> Setting this to 1 may improve the performance of generated code, but may be slower to compile.

I see that you have also removed the "Increase swap file" step - are these related? How does this flag affect memory usage?

aelesbao (approved):
LGTM

.github/workflows/cli.yml
kouks: well, CI passes so it seems it's affected positively. We agreed with @tuky191 that this is good for now as it works and he'll revise this when he has more time

tmsdkeys (approved):
After the discussion yielded no consensus on improved terminology, I agree with current implementation and updated the docs slightly to clarify a bit more

PR #136, opened 2025-04-16
docs: fix links and make syntax consistent

tmsdkeys:
## Notable changes

- Final check before devnet release to fix broken links and make syntax consistent

## References

- Partially closes: https://github.com/Talus-Network/gitbook-docs/issues/46

## Checklist

- [ ] keep a changelog
- [X] create or update documentation

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -8,14 +8,11 @@ The Nexus CLI is a set of tools that is used by almost all Actors in the Nexus e
 - Tool developers use it to scaffold, validate and register tools
 - Nexus developers use it for debugging, testing and both use cases mentioned above
 
-## Open questions
-
-- before public release, Tool<>Leader authorization needs to be implemented
-- publishing and installtion of the CLI, SDK and Tookit needs to be solved
-
 ## Interface design
 
-> Each command can be passed a `--json` flag that will return the output in JSON format. This is useful for programmatic access to the CLI.
+{% hint style="info" %}
+Each command can be passed a `--json` flag that will return the output in JSON format. This is useful for programmatic access to the CLI.
+{% endhint %}
 
 ### `nexus tool`
 
@@ -37,7 +34,9 @@ Validate an off-chain Nexus Tool on the provided URL. This command checks whethe
 2. `GET /health` simple health check endpoint that needs to return a `200 OK` in order for the validation to pass.
 3. `POST /invoke` the CLI can check that the endpoint exists.
 
-> As an improvement, the command could take a `[data]` parameter that invokes the Tool and checks the response against the output schema.
+{% hint style="success" %}
+As an improvement, the command could take a `[data]` parameter that invokes the Tool and checks the response against the output schema.
+{% endhint %}
 
 This command should also check that the URL is accessible by the Leader node. It should, however, be usable with `localhost` Tools for development purposes, printing a warning.
 
@@ -45,7 +44,7 @@ This command should also check that the URL is accessible by the Leader node. It
 
 **`nexus tool validate --on-chain <ident>`**
 
-[Issue #96](https://github.com/Talus-Network/nexus-next/issues/96)
+🚧 This is TBD... 🚧
 
 ---
 
@@ -55,13 +54,15 @@ Command that makes a request to `GET <url>/meta` to fetch the Tool definition an
 
 This returns an OwnerCap object ID that can be used to manage the Tool.
 
-> This command requires that a wallet is connected to the CLI.
+{% hint style="info" %}
+This command requires that a wallet is connected to the CLI..
+{% endhint %}
 
 ---
 
 **`nexus tool register --on-chain <ident>`**
 
-[Issue #96](https://github.com/Talus-Network/nexus-next/issues/96)
+🚧 This is TBD... 🚧
 
 ---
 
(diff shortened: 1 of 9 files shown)


docs/CLI.md
```diff
@@ -37,15 +34,17 @@ Validate an off-chain Nexus Tool on the provided URL. This command checks whethe
...
+{% endhint %}
 
 This command should also check that the URL is accessible by the Leader node. It should, however, be usable with `localhost` Tools for development purposes, printing a warning.
 
 ---
 
 **`nexus tool validate --on-chain <ident>`**
 
-[Issue #96](https://github.com/Talus-Network/nexus-next/issues/96)
+🚧 This is TBD... 🚧
```
aelesbao, line 47: Shouldn't we remove this instead of not documenting it? Reference: https://developers.google.com/style/future

docs/CLI.md
```diff
@@ -55,13 +54,15 @@ Command that makes a request to `GET <url>/meta` to fetch the Tool definition an
...
+{% hint style="info" %}
+This command requires that a wallet is connected to the CLI..
+{% endhint %}
 
 ---
 
 **`nexus tool register --on-chain <ident>`**
 
-[Issue #96](https://github.com/Talus-Network/nexus-next/issues/96)
+🚧 This is TBD... 🚧
```
aelesbao, line 65: Same here

docs/guides/setup.md
```diff
@@ -25,9 +25,7 @@ brew install nexus-cli
 
 #### Arch Linux
 
-The [nexus-cli](https://aur.archlinux.org/packages/nexus-cli) is also available
-in the AUR (Arch User Repository). You can install it using your preferred
-[AUR helper](https://wiki.archlinux.org/title/AUR_helpers):
+The [nexus-cli](https://aur.archlinux.org/packages/nexus-cli) is also available in the AUR (Arch User Repository). You can install it using your preferred [AUR helper](https://wiki.archlinux.org/title/AUR_helpers):
```
aelesbao, line 28: We should setup a linter later, but it's recommended to [keep a maximum line length of 80 characters for markdown files](https://google.github.io/styleguide/docguide/style.html#character-line-limit) since not everyone works with word wrap enabled in their editors.

docs/CLI.md
tmsdkeys: The counter argument is that the CLI already does reference it. So the user could be confused

kouks (approved):

davidrotari19 (approved):

docs/guides/setup.md
kouks: I agree with @tmsdkeys here, I think a paragraph should be on a single line in `.md` files. If we cut it to 80 characters, the rendering can become awkward because we're unlikely to use a monospace font in the docs. Also resizing the screen will mess up the width also.

tmsdkeys: https://www.perplexity.ai/search/when-using-gitbook-for-docs-sh-BSwmKsjPTx2Ou.YrBcETCg#0

Should be take care of by Gitbook

docs/CLI.md
```diff
@@ -83,15 +84,17 @@ After the period of time configured in our Tool Registry, let the Tool developer
...
+This command requires that a wallet is connected to the CLI..
+{% endhint %}
 
 **`nexus tool list`**
 
 List all Nexus Tools available in the Tool Registry. This reads the dynamic object directly from Sui.
 
-> This command requires that a wallet is connected to the CLI.
+{% hint style="info" %}
+This command requires that a wallet is connected to the CLI..
```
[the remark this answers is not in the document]
tmsdkeys: fixed in https://github.com/Talus-Network/nexus-sdk/pull/136/commits/f99a64036c4626a7ab1b580449be2bcb73180619

docs/CLI.md
aelesbao: I agree with @kouks here, we should let the users know this command isn't implemented yet.

docs/guides/setup.md
tmsdkeys: @aelesbao it does look off though in gitbook: https://docs.talus.network/getting-started/setup#access-devnet-sui-explorer or https://docs.talus.network/getting-started/setup#arch-linux

kouks: Fair enough, in that case I don't have a strong opinion here. ctrl+shift+p "word wrap" is imo easy enough to trigger in vscode. But if we have a linter that formats automatically I don't mind either way.

I'm all for linting for consistency in the docs but something post-devnet.

docs/CLI.md
tmsdkeys: will rephrase then

docs/guides/setup.md
kouks: > Fair enough, in that case I don't have a strong opinion here. ctrl+shift+p "word wrap" is imo easy enough to trigger in vscode. But if we have a linter that formats automatically I don't mind either way.

Scrape this. No newlines then.

docs/CLI.md
tmsdkeys: rephrased in [4f6b9a0](https://github.com/Talus-Network/nexus-sdk/pull/136/commits/4f6b9a079f06a2740aeaaa489dbdd22346192cc3)

docs/CLI.md
tmsdkeys: rephrased in [4f6b9a0](https://github.com/Talus-Network/nexus-sdk/pull/136/commits/4f6b9a079f06a2740aeaaa489dbdd22346192cc3)

PR #150, opened 2025-04-29
Feat/tool walrus tool 2.endpoint is read json

afarukcali:
## Notable changes

- This pull request covers the JSON read functionality within the Walrus tool. Users can obtain their outputs in both JSON format and raw string format.

## Checklist

- [x] keep a changelog
- [x] write tests
- [ ] ~~create or update documentation~~


tools/walrus/src/json/read_json.rs
```diff
@@ -0,0 +1,264 @@
...
+    #[serde(default)]
+    aggregator_url: Option<String>,
+}
+
+#[derive(Serialize, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub(crate) enum Output {
+    Ok { json: Value, text: String },
+    Err { reason: String },
+}
```
kouks, line 40: if you're making tools with generic JSON output, the JSON structure should be hinted and validated against a schema that's an input of the tool akin to: https://github.com/Talus-Network/nexus-sdk/blob/main/tools/llm-openai-chat-completion/src/main.rs#L215

tools/walrus/src/json/read_json.rs
```diff
@@ -0,0 +1,264 @@
...
+impl NexusTool for ReadJson {
+    type Input = Input;
+    type Output = Output;
+
+    async fn new() -> Self {
+        Self {}
+    }
+
+    fn fqn() -> ToolFqn {
+        fqn!("xyz.taluslabs.walrus.json.read@1")
```
kouks, line 53: not sure if you've not got around to updating these but similarly this should be `xyz.taluslabs.storage.walrus.json-read`

afarukcali: https://github.com/Talus-Network/nexus-sdk/pull/150/commits/6d46950630d96d67a4688430dcd33c1dc9919967 fixed.

tools/walrus/src/json/read_json.rs
kouks: We don't need to design the schema, we want to allow the user of the tool to pass it and validate the output against it. You can basically copy the LLM tool functionality (the line I linked). Think of it as the tool being generic over `T` and the user passes the structure of `T` as an input. And by validating against the schema, we ensure that the output is indeed `T`. This is also gonna be very useful for GUI.

afarukcali: https://github.com/Talus-Network/nexus-sdk/pull/150/commits/0929a2a511afb6280bb56846a8cffef50f280630 I tried to adjust it as you suggested, now I’m waiting for your feedback.

kouks: I believe json schema has a validate function, you can refer to the LLM tool again, no need to write that huge recursive function

kouks (approved):

tools/storage-walrus/src/read_json.rs
```diff
@@ -0,0 +1,610 @@
...
+
+        Ok(storage_info)
+    }
+}
+
+// Validation utility struct and function to use all schema fields
+struct ValidationResult {
+    is_valid: bool,
+    error_message: String,
+}
```
aelesbao, line 196: The `ValidationResult` struct introduces ambiguity - it's possible to have `is_valid` set to true while still containing an error message, which can be misleading.

A more idiomatic approach is to have `create_validation_configuration` return a `Result<(), ReadJsonError>`, so you avoid some of the boolean checks you have later.

aelesbao (changes requested):

tools/storage-walrus/src/read_json.rs
```diff
@@ -0,0 +1,610 @@
...
+
+fn create_validation_configuration(
```
aelesbao, line 198: This function is actually validating the JSON, so it'd be better to rename it to simply `validate` instead.

afarukcali: https://github.com/Talus-Network/nexus-sdk/pull/150/commits/d75dd4d0e5fc1c309090766d1c091fc9cc5a98c3 renamed.

aelesbao (approved):

PR #188, opened 2025-05-23
294 add crypto conf key to store the leaders public key and the current shared secret and index

davidrotari19:
## Notable changes

## References

## Checklist

- [ ] keep a changelog
- [ ] write tests


cli/src/prelude.rs
```diff
@@ -43,6 +43,8 @@ pub(crate) struct CliConf {
     pub(crate) nexus: Option<NexusObjects>,
     #[serde(default)]
     pub(crate) tools: HashMap<ToolFqn, ToolOwnerCaps>,
+    #[serde(default)]
+    pub(crate) state_store: StateStoreConf,
```
kouks, line 47: The naming here should be more explicit about it being crypto state. At the moment it's too generic imo.

sdk/src/crypto/state_store.rs
```diff
@@ -0,0 +1,458 @@
...
+
+    fn try_from(stored: StoredIdentityKey) -> Result<Self, Self::Error> {
+        let secret: StaticSecret = stored.secret.into();
+        Ok(IdentityKey::from_secret(secret))
+    }
+}
+
+// Session is stored in the state store
+#[derive(Serialize, Deserialize)]
+struct StoredSession {
```
kouks, line 84: I'm not sure if this belongs to the `crypto` module. I think `Session` should just be `Serialize, Deserialize` and we handle the storing in the CLI and the leader because the storing methods are quite different.

Also - are we writing files to the disk and then referencing them in the config? Why not just store the bytes in the config file directly?

davidrotari19: Thats a good question. The way I looked at it we store, secret keys(several of them). So maybe we should encrypt them and just store the path, to the encrypted file. Or we could just do the encrypted in the file directly. I probably overdone this

davidrotari19: If this is just too much, I will just revert to `Serialize, Deserialize`

cli/src/prelude.rs
davidrotari19: Maybe `SecureStateConf` or `SecureSessionConf`?

cli/src/prelude.rs
```diff
@@ -98,6 +100,20 @@ impl Default for SuiConf {
...
+    pub(crate) state_file: std::path::PathBuf,
+}
+
+fn default_state_path() -> std::path::PathBuf {
+    home::home_dir()
+        .unwrap()
+        .join(".nexus")
+        .join("state_store.bin")
+}
+
```
aelesbao, line 116: We should prefer using the `XDG_CONFIG_HOME` path [as a standard](https://specifications.freedesktop.org/basedir-spec/latest/) instead and fallback to `HOME` only if this env var isn't availble.

cli/src/prelude.rs
kouks: or just `crypto`

davidrotari19: [27db7b0](https://github.com/Talus-Network/nexus-sdk/pull/188/commits/27db7b060b3a6702d7e3c8ab2a9fe4da15e912dd)

cli/src/prelude.rs
davidrotari19: [27db7b0](https://github.com/Talus-Network/nexus-sdk/pull/188/commits/27db7b060b3a6702d7e3c8ab2a9fe4da15e912dd) I removed this completely now we save state directly

cli/src/prelude.rs
```diff
@@ -37,14 +39,14 @@ impl std::fmt::Display for SuiNet {
 }
 
 /// Struct holding the config structure.
-#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
```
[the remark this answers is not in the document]
kouks: that's fine configs are usually fine to pass by reference

cli/Cargo.toml
```diff
@@ -35,6 +35,19 @@ thiserror.workspace = true
...
+rand = "0.8"
+# Encryption helpers
+aes-gcm = "0.10"
+base64 = "0.21"
+bincode = "1"
+argon2 = "0.5"
+directories = "5"
+hex = "0.4"
+keyring = "3.6"
+zeroize = "1.7"
```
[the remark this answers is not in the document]
davidrotari19: [86020a2](https://github.com/Talus-Network/nexus-sdk/pull/188/commits/86020a208772ede5d5156cf0ad87b2b54f3352fb)

kouks (approved):

PR #213, opened 2025-06-16
chore(leader): configure cargo-deny rules

aelesbao:
## Notable changes

- added rules for dependencies licenses in `deny.toml`

## Checklist

- [x] keep a changelog
- [ ] ~write tests~
- [ ] ~create tickets for `TODO: <>` statements~
- [ ] ~create or update documentation~


deny.toml
```diff
@@ -0,0 +1,243 @@
+# This template contains all of the possible sections and their default values
+
+# Note that all fields that take a lint level have these possible values:
```
davidrotari19, line 1: Do we need to keep all the text here?

aelesbao: The first line is irrelevant, but I'd prefer to keep the other comments as a reference to anyone who's editing the file later.

aelesbao: 2e0cacdda99e01577b8e1ebcbf52dd604d27f799

PR #215, opened 2025-06-18
fix: Prevent URL scheme injection attacks in meta_handler

aso20455:
## Description
This PR adds validation for the X-Forwarded-Proto header in the meta_handler function to prevent potential URL injection attacks. The change ensures that only 'http' and 'https' schemes are allowed, rejecting any potentially malicious schemes that could lead to security vulnerabilities.

## Problem
Currently, the meta_handler function accepts any value for the X-Forwarded-Proto header without validation. This could allow an attacker to inject malicious URL schemes (like javascript:, data:, etc.) which might lead to XSS or other security issues.

## Solution
Added explicit validation to ensure the scheme is either 'http' or 'https', returning a 400 Bad Request with a clear error message if an invalid scheme is provided.

## Checklist

- [ ] keep a changelog
- [ ] write tests

### DIFF
--- a/toolkit-rust/tests/integration.rs
+++ b/toolkit-rust/tests/integration.rs
@@ -235,13 +235,13 @@ mod tests {
 
     #[tokio::test]
     async fn test_meta_invalid_schema() {
-        tokio::spawn(async move { bootstrap!(([127, 0, 0, 1], 8043), DummyTool) });
+        tokio::spawn(async move { bootstrap!(([127, 0, 0, 1], 8047), DummyTool) });
 
         // Give the webserver some time to start.
         tokio::time::sleep(std::time::Duration::from_millis(100)).await;
 
         let meta = Client::new()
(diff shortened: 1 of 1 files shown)


toolkit-rust/src/runtime.rs
```diff
@@ -251,7 +265,7 @@ async fn invoke_handler<T: NexusTool>(input: serde_json::Value) -> Result<impl R
         Err(e) => {
             let reply = json!({
                 "error": "input_deserialization_error",
-                "details": e.to_string(),
+                "details": format!("Failed to parse input: {}", e.to_string().split('\n').next().unwrap_or_default())
```
kouks: is this change necessary? what is the motivation behind it?

aso20455: @kouks This line formats the error message by adding a "Failed to parse input:" prefix and extracting only the first line of the original error to improve API response readability.

PR #222, opened 2025-06-23
fix nightly version, since nightly is a moving target

loverdos:
See also: https://github.com/Talus-Network/nexus-next/pull/368


loverdos:
OK, apparently there is [more] to it, I'll continue later or tomorrow.

[more]: https://github.com/Talus-Network/nexus-sdk/actions/runs/15828604724/job/44615251035?pr=222

PR #224, opened 2025-06-24
Reduce the size of error enums

c410-f3r:
## Notable changes

A large enum used in several places can cause a negative runtime impact. In fact, this is so common that the community created several lints to prevent such a scenario.

- [large_enum_variant](https://rust-lang.github.io/rust-clippy/master/?groups=perf#large_enum_variant)
- [result_large_err](https://rust-lang.github.io/rust-clippy/master/?groups=perf#result_large_err)
- [variant_size_differences](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_lint/types/static.VARIANT_SIZE_DIFFERENCES.html)

This PR cuts 8~24 bytes from different error enums through the use of `Box<str>`, which doesn't change any intended behavior.

Of course, 8~24 bytes won't make much of a runtime different in this project but it is still worthwhile IMO. 

## Checklist

- [ ] keep a changelog
- [ ] write tests


davidrotari19:
Nice, thank you for this. Could you please fix the issues and check the boxes in the template.

Issue #234, opened 2025-07-08
Set up a pre-commit hook


loverdos:
I have observed that many times CI checks fail for silly reasons (one wrongly formatted `use` statement in Rust code), while we could have prevented that by doing some checks locally before committing or pushing. 

Of course one could argue these are all fixable by properly setting up the editor/IDE but in practice it's not reliable.

Whether to do the pre-commit or the pre-push hook is debatable, I'll go for the former.

PR #235, opened 2025-07-08
Set up a pre-commit hook

loverdos:
Run `just pre-commit` to see a nice list of what we run. Currently it looks like this:

```sh
Available recipes:
    cargo-check             # cargo check --locked --workspace --bins --examples
    cargo-clippy            # cargo clippy --all-targets --all-features
    cargo-clippy-strict     # cargo clippy --all-targets --all-features -- -D warnings
    cargo-nextest-build     # cargo nextest run --locked --no-run --workspace --all-features --bins --examples --tests
    cargo-nextest-run       # cargo nextest run --locked --fail-fast --workspace --all-features --bins --examples --tests
    cargo-nightly-fmt       # cargo +nightly-2025-01-06 fmt --all --check
    check-toolchain-version # cargo_version == toolchain_version
    md-lint                 # markdownlint-cli2 '**/*.md' '!.git/' --config .markdownlint.json
```

All these, except `cargo-clippy-strict` which will require more work and consensus, are nicely orchestrated by https://github.com/loverdos/pre-commit 😄 inside the `.pre-commit` top-level folder.

**How to run all these**
Run the hook directly, which is what `git` `.git/hooks/pre-commit` points to:

```sh
./.pre-commit/pre-commit
[...]
== Results: 7 success(es), 0 failure(s)
== ✅ .pre-commit/00-toolchain-version
== ✅ .pre-commit/01-cargo-check
== ✅ .pre-commit/02-cargo-clippy
== ✅ .pre-commit/03-cargo-nightly-fmt
== ✅ .pre-commit/04-cargo-nextest-build
== ✅ .pre-commit/05-cargo-nextest-run
== ✅ .pre-commit/06-md-lint
==
```

**How do I set up the hook locally?**
I describe this in CONTRIBUTING.md but here is the recipe:

```
./.pre-commit/pre-commit --install
```

**How did you install it the first time, before it even existed in the repo?**
I followed the instructions from https://github.com/loverdos/pre-commit.

## Notable changes

- see ticket and changelog

## References

- https://github.com/Talus-Network/nexus-sdk/issues/234

## Checklist

- [x] keep a changelog
- [x] write tests (`pre-commit` actions actually)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - CONTRIBUTING.md
 - CODE_OF_CONDUCT.md
+- `pre-commit` hook
 
 ### `nexus-cli`
 


docs/guides/math-branching-with-chat.md
```diff
@@ -3,9 +3,9 @@
 This guide builds on the [Math Branching DAG with Entry Groups][math-branching-entry-guide] by adding a chat completion tool that explains the mathematical results. You'll learn how to:
 
 1. Understand the need for a custom tool to bridge between math operations and chat completion
-2. Add the chat completion tool to the DAG
-3. Connect the math results to the chat completion tool
-4. Configure the chat completion tool for optimal results
+1. Add the chat completion tool to the DAG
+1. Connect the math results to the chat completion tool
```
davidrotari19, line 7: why are they all one?

.pre-commit/.just
```diff
@@ -0,0 +1,74 @@
...
+cargo-nightly-fmt: _check-cargo
+    #!/usr/bin/env bash
+
+    cargo +nightly-2025-01-06 fmt --all --check || {
+        echo "== Formatting failed. Please run 'cargo +nightly fmt --all' to format the code." >&2
+        exit 1
+    }
+
+# cargo clippy --all-targets --all-features
+cargo-clippy: _check-cargo
```
davidrotari19, line 44: This is duplicate with the one  `cargo-clippy-strict`, should we pass a flag?

.pre-commit/pre-commit
```diff
@@ -0,0 +1,197 @@
...
+    cd "$git_root"
+
+    # Make sure all the needed folders are there
+    mkdir -p $dot_pre_commit
+    mkdir -p $dot_git_hooks
+
+    if is-it-me; then
+        # We are just developing ourselves
+        LOG "Hey, I noticed you are messing around with me :)"
+        cp $pre_commit $dot_pre_commit_pre_commit
```
davidrotari19, line 85: The expansions need to be in quote, it will not work in case of spaces. May not be a problem here, but can cause problems in general.

docs/guides/math-branching-with-chat.md
loverdos: So I discovered we have a markdown linting [rule] that requires exactly that. In fact, I also discovered that most of the markdown files on `main` do not get past this rule (we missed them somehow), hence the corrections here. The exact code for the rule is [in our config][MD029-config] and is defined [here][MD029].


[rule]: https://github.com/Talus-Network/nexus-sdk/blob/a9f6ed643c1b03094819984938ad357f6e596fc9/.github/workflows/markdown_format.yml#L28

[MD029-config]: https://github.com/Talus-Network/nexus-sdk/blob/a9f6ed643c1b03094819984938ad357f6e596fc9/.markdownlint.json#L10

[MD029]: https://github.com/DavidAnson/markdownlint/blob/main/doc/md029.md

.pre-commit/pre-commit
loverdos: [`shellcheck`](https://github.com/loverdos/pre-commit/blob/fdfb0a7b38e7175f45de69c2bd92eac2e342805e/.pre-commit/00-shellcheck#L3) does not complain, so it should be fine. In fact if you notice the values of these variable, they are space-free.

This file is maintained elsewhere.

.pre-commit/.just
loverdos: Good question. I opted being lazy :-D and copy-pasted.

davidrotari19: I do not see a problem, it was more like an observation.

docs/guides/math-branching-with-chat.md
davidrotari19: Seems unnatural to me, should we move to ordered?

loverdos: I would agree but here I am just enforcing the rules we have chosen, and as I said, according to our rules `main` is broken. Let's do it in another PR, feel free to open a ticket under https://github.com/Talus-Network/nexus-sdk/issues/236

davidrotari19: https://github.com/Talus-Network/nexus-sdk/issues/237

loverdos:
> is MY_URL currently a potential security threat? Because the script seems to fetch code from there and execute on our machines.

`MY_URL` is basically the `URL` used in the recipe:

```
bash -c "bash <(curl -fsSL https://raw.githubusercontent.com/loverdos/pre-commit/main/pre-commit) --install"
```

which is the script's [installation manual](https://github.com/loverdos/pre-commit/blob/fdfb0a7b38e7175f45de69c2bd92eac2e342805e/pre-commit#L27). I guess it's all about "do you trust a random shell installation script from the internet"? In our case it will become random if someone changes `MY_URL`. If you do and then run `.prec-commit/pre-commit --install`, then the script will access the new `MY_URL`.

Issue #254, opened 2025-07-17
Add `pre-commit` check for docker


loverdos:
Do we need to create an issue for this?

_Originally posted by @nonast in https://github.com/Talus-Network/nexus-sdk/pull/241#discussion_r2212300657_

PR #261, opened 2025-07-22
feat: add pre-commit check for docker

loverdos:
Ref: https://github.com/Talus-Network/nexus-sdk/issues/254


.pre-commit/.just
```diff
@@ -19,6 +19,16 @@ cargo-nightly-fmt: _check-cargo
...
     }
 
+
+check-docker: _check-docker
+    #!/usr/bin/env bash
+
+    if ! docker info > /dev/null 2>&1; then
+        echo "❌ Docker is not running. Please start Docker and try again." >&2
+        exit 1
+    fi
```
aelesbao: IMHO, we shouldn't enforce developers to use Docker when there are multiple Docker-compatible options available (ex: `podman`). With podman, for example, it's possible to configure the `DOCKER_HOST` in a way that it's transparent for `testcontainers` where it's running the containers.

just/_helpers.just
```diff
@@ -10,6 +10,17 @@ _check-cargo:
...
+_check-docker:
+    #!/usr/bin/env bash
+
+    if ! command -v docker &> /dev/null
+    then
+        echo "docker could not be found"
+        echo "Install docker from https://docs.docker.com/get-docker/"
+        exit 1
+    fi
+
```
aelesbao: Same as [here](https://github.com/Talus-Network/nexus-sdk/pull/261/files#r2222517853).

.pre-commit/.just
loverdos: Ah! yes that's a good point. So I need to figure out a version of these checks that is compatible with all the other software.

Let me dive in.

just/_helpers.just
loverdos: Code is now able to handle any runtime, currently also supporting the checks for `podman` but it's extensible, see https://github.com/Talus-Network/nexus-sdk/pull/261/commits/e0667abc4ede26bf98e57dad5764bb711c54b6a1

Unfortunately, see https://github.com/Talus-Network/nexus-sdk/pull/261/commits/e0667abc4ede26bf98e57dad5764bb711c54b6a1

So I guess there needs to be follow up work, for now Docker is good enough I think.

Issue #262, opened 2025-07-23
Check that the test code is container runtime agnostic


loverdos:
_Original context in https://github.com/Talus-Network/nexus-sdk/pull/261#discussion_r2222517853_

aelesbao:
This is a limitation of `testcontainers`. In my computer I had to create a symlink with [the path it expects the Docker socket to be](https://docs.rs/testcontainers/0.24.0/testcontainers/#the-host-is-resolved-in-the-following-order). Ex:

```sh
ln -s ${XDG_RUNTIME_DIR}/podman/podman.sock ${HOME}/.docker/run/docker.sock
```

loverdos:
I tested on macOS/M2 and this works with `podman` running:

```
export DOCKER_HOST=unix://$HOME/.local/share/containers/podman/machine/podman.sock
```

PR #265, opened 2025-07-25
Reduce code duplication in move object deserialization (object crawler)

loverdos:
Ref: https://github.com/Talus-Network/nexus-sdk/issues/264

The changes are straightforward, and indeed that was what I had in mind when I was thinking about making such changes but this took a bit longer because I stumbled upon a bug (object crawler test failing) that now I cannot reproduce ... 🤷


.pre-commit/.just
```diff
@@ -40,6 +40,12 @@ check-container-runtime:
         [podman]="podman info > /dev/null 2>&1"
     )
 
+    # If DOCKER_HOST is set print it.
+    # Sometimes it has the wrong value, so it's good to know.
+    if [[ -n "${DOCKER_HOST:-}" ]]; then
+        echo "DOCKER_HOST=$DOCKER_HOST"
+    fi
+
```
loverdos, line 48: Not relevant to this PR, just a good opportunity to sneak this in.

Based on a true story -- I had my DOCKER_HOST set incorrectly, basically forgotten from the https://github.com/Talus-Network/nexus-sdk/issues/262 investigation.

sdk/Cargo.toml
```diff
@@ -217,7 +217,7 @@ optional = true
 
 [dependencies.tokio]
 workspace = true
-features = ["io-util"]
+features = ["io-util"] # include "macros" to run #[tokio::test] from the IDE
```
aelesbao: Shouldn't we add it then instead of leaving a comment?

loverdos: I had it locally but I am not sure if it's useful to everybody. Perhaps it increases binary size, but I am not sure. If it was just me, or next time I will have to use this feature, I would just add it.

kouks: add it as dev dependencies if it's just for running test from the IDE

kouks (approved):

davidrotari19 (approved):

sdk/Cargo.toml
loverdos: @kouks https://github.com/Talus-Network/nexus-sdk/pull/265/commits/81b553285e6607791413d6593c1f71d3ac84c30a

Issue #257, opened 2025-07-17
Sync the `pre-commit` checks with what CI is doing.


loverdos:
The idea is to have one source of truth for some checks, and make sure they all run and in the same way. 

Example: CI was happy to merge https://github.com/Talus-Network/nexus-sdk/pull/256 but apparently:

```sh
$ ./.pre-commit/pre-commit
[...]
==
== Results: 8 success(es), 1 failure(s)
== ✅ .pre-commit/01-cargo-check
== ✅ .pre-commit/02-cargo-clippy
== ✅ .pre-commit/03-cargo-nightly-fmt
== ✅ .pre-commit/04-cargo-nextest-build
== ✅ .pre-commit/05-cargo-nextest-run
== ❌ .pre-commit/06-md-lint
== ✅ .pre-commit/07-check-shebang
== ✅ .pre-commit/08-check-just-shebang
== ✅ .pre-commit/09-typos
==
```

See also the [Slack thread](https://taluslabs.slack.com/archives/C06EFET91KR/p1752763810262899?thread_ts=1752689152.481709&cid=C06EFET91KR).

loverdos:
When the cache works, timings looks like this:

```
==
== Results: 12 success(es), 0 failure(s)
==      In: 03:42 [min:sec]
== 
== ✅ [00:14] .pre-commit/01-cargo-check
== ✅ [00:41] .pre-commit/02-cargo-clippy
== ✅ [00:01] .pre-commit/03-cargo-nightly-fmt
== ✅ [00:03] .pre-commit/04-check-container-runtime
== ✅ [00:57] .pre-commit/05-cargo-nextest-build
== ✅ [01:22] .pre-commit/06-cargo-nextest-run
== ✅ [00:01] .pre-commit/07-md-lint
== ✅ [00:00] .pre-commit/08-check-shebang
== ✅ [00:00] .pre-commit/09-check-just-shebang
== ✅ [00:00] .pre-commit/10-typos
== ✅ [00:22] .pre-commit/11-test-cli-completions
== ✅ [00:01] .pre-commit/12-executable-scripts
==
```
`03:42` is much better than current (pre-this-PR) timings, which are around 13+ minutes.
Check `11` does a full build of the `cli` so it is indicative of how much it takes for a full build of anything with caching enabled.

It is interesting to note that with what we have in main currently, caching does not work. Perhaps the interaction with `CARGO_INCREMENTAL` creates problems.
[...]

PR #267, opened 2025-08-01
Sync the pre-commit checks with what CI is doing

loverdos:
## Description and remarks

- Run the pre-commit hook in CI. It basically supersedes most other tests, hence some deletions of `.github/workflows/sdk.yml` etc.
- Run times are from 6-8 minutes to 24-26 minutes depending on a) if the cache kicks in, b) if GitHub is having a good day and does not fail cache saving.
- When the cache works, building and testing everything is faster than selectively building/testing individual crates, not to mention that we always get the dependencies right.
- Add one pre-commit action to handle a CLI integration test (for CLI completions in bash)
- Add one pre-commit check for executable scripts. We are now being strict, in the sense if something looks like a shell script, it must be executable.
- Introduce a two-layered CI cache for a) the `main` branch, b) PRs. Both interoperate.
- Unfortunately, the caching system of GitHub is unreliable. Sometimes caches are saved, sometimes they are not. Still, anecdotally we are doing better than the current (previous to this PR) setup, where we observe that generally cache does not work the way we use it.
- I have come to the conclusion that it's better to use `actions/cache` manually, rather than relying on marketplace GitHub actions, such as `actions-rust-lang/setup-rust-toolchain` or `Swatinem/rust-cache`. Generally, GitHub actions are not composable.
- At some point, when everything seemed stable, I introduced `sccache`. It looked promising, until I started getting weird errors out of the blue, like `warning: ring@0.17.14: sccache: error: failed to create directory ~/.cache/sccache/preprocessor`, and so I abandoned `sccache` altogether. Life is short.
- In general, though this was an interesting learning exercise, it has mostly been full of frustrations. I am certain I did not get everything right and there is room for improvement, but I prefer we move to another system/way of doing CI than spend time fighting GitHub actions. Honestly, good old Jenkins on a vanilla VM would be better/more reliable (but I am not saying we should do that).
- I tried several variations of making reusable GitHub actions, but I decided to stop here. Please do not recommend more in your review. Feel free to open a ticket to address this in the future.
- There are some other workflows that overlap, e.g. [checking for typos](https://github.com/Talus-Network/nexus-sdk/blob/81985d2648081737aa9bf780f49b013bc05d380c/.github/workflows/markdown_format.yml#L48). TODO https://github.com/Talus-Network/nexus-sdk/issues/272
- All of these notes will go to the commit message.

## References

- https://github.com/Talus-Network/nexus-sdk/issues/257

## Checklist

- [x] keep a changelog
- [ ] ~~write tests~~
- [ ] ~~create or update documentation~~


.pre-commit/pre-commit
loverdos, line 1: I updated this file manually from its source repo.

helpers/helpers.just
```diff
@@ -0,0 +1,115 @@
...
+    exe=typos
+    if ! command -v $exe &> /dev/null
+    then
+        echo "$exe could not be found"
+        echo "Install $exe from https://github.com/crate-ci/typos"
+        exit 1
+    fi
+
+[private]
+detect-changes +GLOBS:
```
loverdos, line 49: This ended up not being used, so I guess I can just delete it.

loverdos: Nice utility though, I like it :)

aelesbao (approved):

helpers/helpers.just
zanicar: If it's useful, don't delete it without archiving or documenting it. It may be relevant in the future, who knows...

zanicar (approved):
LGTM

PR #268, opened 2025-08-04
Draft: [0.3] Add support for looping

kouks:
## Notable changes

## References

## Checklist

- [ ] keep a changelog
- [ ] write tests

### DIFF
--- a/sdk/src/types/json_dag.rs
+++ b/sdk/src/types/json_dag.rs
@@ -79,6 +79,24 @@ pub enum Data {
 pub struct Edge {
     pub from: FromPort,
     pub to: ToPort,
+    /// The kind of the edge. This is used to determine how the edge is
+    /// processed in the workflow. Defaults to [`EdgeKind::Normal`].
+    #[serde(default)]
+    pub kind: EdgeKind,
+}
+
+// TODO: validator changes and tests
+#[derive(Clone, Debug, Default, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum EdgeKind {
+    #[default]
+    Normal,
+    /// For-each and collect control edges.
+    ForEach,
+    Collect,
+    /// Do-while and break control edges.
+    DoWhile,
+    Break,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Deserialize)]
@@ -88,6 +106,7 @@ pub struct FromPort {
     pub output_port: String,
     /// Whether the output port data should be encrypted before being sent to
     /// the workflow. Defaults to `false`.
+    // TODO: maybe move this to `Edge`?
     #[serde(default)]
     pub encrypted: bool,
 }
--- a/sdk/src/events.rs
+++ b/sdk/src/events.rs
@@ -84,7 +84,7 @@ pub struct RequestWalkExecutionEvent {
         serialize_with = "serialize_sui_u64"
     )]
     pub walk_index: u64,
-    pub next_vertex: TypeName,
+    pub next_vertex: RuntimeVertex,
     pub evaluations: sui::ObjectID,
     /// This field defines the package ID, module and name of the Agent that
     /// holds the DAG. Used to confirm the tool evaluation with the Agent.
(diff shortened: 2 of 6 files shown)


docs/guides/dag-construction.md
```diff
@@ -75,14 +75,25 @@ Edges define the flow of data between vertices, connecting an output port of a s
...
 ```
 
 - The `source_vertex_name` and `target_vertex_name` refer to the `name` field of vertices defined in the `vertices` list.
 - The `target_input_port_name` must be a valid input port for the tool used by the `target_vertex_name`.
 - If the `encrypted` field is set to `true`, the data will be encrypted before being sent on-chain. This is useful for sensitive data that should not be exposed in plaintext.
 
+## 3.1 Special edge types
+
+- **`normal`** → The default edge type. Data flows from the source vertex to the target vertex as soon as the source vertex produces output on the specified port.
+- **`for_each`** → Spawns parallel executions of the downstream vertex, once per element of an array.
```
davidrotari19: It may need clarification that the next edge type of the downstream vertex has to be collect.

davidrotari19 (approved):

sdk/src/dag/validator.rs
```diff
@@ -195,6 +207,133 @@ fn get_net_concurrency_in_subgraph(
...
+
+                (InForEach, EdgeKind::Normal) => InForEach,
+                (InForEach, EdgeKind::ForEach) => {
+                    bail!("'{vertex}' has a nested for-each");
+                }
+                (InForEach, EdgeKind::Collect) => Idle,
+                (InForEach, EdgeKind::DoWhile) | (InForEach, EdgeKind::Break) => {
+                    bail!("'{vertex}' has a do-while or break edge inside a for-each");
+                }
+            };
```
aelesbao, line 265: Nice, easy to reason about :+1:

sdk/src/idents/primitives.rs
```diff
@@ -51,6 +54,51 @@ impl Data {
         json: &T,
         encrypted: bool,
     ) -> anyhow::Result<sui::Argument> {
+        // Arrays need to be created with `Self::INLINE_MANY`.
```
aelesbao, line 57: I think it'd be better to break this fn down into separate functions. It currently has 4 exit points, which increases complexity. One quick change I see is to remove the repeated code block that checks for encryption. Ex:

```rust
fn build_programmable_move_call(
  mod_name: ModuleAndNameIdent,
  encrypted_mod_name: ModuleAndNameIdent,
  arguments: Vec<Argument>,
) {
  if encrypted {
      return Ok(tx.programmable_move_call(
          primitives_pkg_id,
          encrypted_mod_name.module.into(),
          encrypted_mod_name.name.into(),
          vec![],
          arguments,
      ));
  }
  
  Ok(tx.programmable_move_call(
      primitives_pkg_id,
      mod_name.module.into(),
      mod_name.name.into(),
      vec![],
      arguments,
  ))
}
```

sdk/src/types/json_dag.rs
```diff
@@ -88,6 +105,7 @@ pub struct FromPort {
     pub output_port: String,
     /// Whether the output port data should be encrypted before being sent to
     /// the workflow. Defaults to `false`.
+    // TODO: maybe move this to `Edge`?
```
aelesbao: In this case, would you keep both levels of encryption (Edge and Port)?

kouks: Not sure what exactly you mean here but this would change nothing functionality wise. It would only "improve" the structure slightly. But since encryption is a finished feature and this wouldn't be the smallest change, I'll probably just remove the comment in favour of a ticket.

aelesbao: From your comment, I understand that you are considering moving the encryption flag from the port to the entire edge (i.e. all ports on it). I'm just wondering if it would make sense to keep both, so that people could decide granularly what's encrypted.

kouks: https://github.com/Talus-Network/nexus/issues/524

kouks: There is no difference in what is encrypted between those 2.

docs/guides/dag-construction.md
kouks: https://github.com/Talus-Network/nexus-sdk/pull/268/commits/79243d7400b54756873fde4c9f01b8e1eefa62f2

sdk/src/idents/primitives.rs
kouks: I tried this and in my opinion it's less readable than leaving it like this. I think the function overall makes sense (4 results - encrypted one, encrypted many, plain one, plain many).

PR #281, opened 2025-08-13
fix/crypto-init-bug

davidrotari19:
## Notable changes

## References

## Checklist

- [ ] keep a changelog
- [ ] write tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [`0.3.0`] - unreleased 
+
+### Repository
+
+### `nexus-cli`
+
+
 ## [`0.2.0`] - 2025-08-12
 
 ### Repository
--- a/cli/src/crypto/crypto_init_key.rs
+++ b/cli/src/crypto/crypto_init_key.rs
@@ -11,6 +11,7 @@ use {
 };
 
 /// Generate and store a new 32-byte key in the OS key-ring.
+/// This will also wipe any crypto configuration from the CLI configuration file.
 pub async fn crypto_init_key(force: bool) -> AnyResult<(), NexusCliError> {
     command_title!("Generating and storing a new 32-byte master key");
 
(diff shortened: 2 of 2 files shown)


.pre-commit/.just
```diff
@@ -34,34 +32,21 @@ check-container-runtime:
     # investigation, so:
     # NOTE: https://github.com/Talus-Network/nexus-sdk/issues/262
 
-    runtimes=(docker podman) # colima etc ...
-    check_commands=(
-        [docker]="docker info > /dev/null 2>&1"
-        [podman]="podman info > /dev/null 2>&1"
-    )
+    runtimes="docker podman" # colima etc ...
```
loverdos: I understand you are refactoring because the command called is "info" in both cases but the original code is general enough for _every_ case while adding minimal cost (source code lines, understandability, maintenance), so I am opposing to this change.

davidrotari19: If you mean the whole receipt, then for me it did not work. It used associative vectors, which are not supported by the default bash used by MacOS. And adding a new dependency on Bash did not seem worth it to me; the check can be added to POSIX too.

davidrotari19: Reverted back [7f99aa0](https://github.com/Talus-Network/nexus-sdk/pull/281/commits/7f99aa04529fc6b17a606a10fae023f2236711a8)

cli/src/crypto/crypto_init_key.rs
```diff
@@ -33,7 +34,65 @@ pub async fn crypto_init_key(force: bool) -> AnyResult<(), NexusCliError> {
...
+            cleanup_handle.success();
+        }
+        Err(e) => {
+            cleanup_handle.error();
+            return Err(NexusCliError::Any(anyhow!(
+                "Failed to check existence of CLI configuration: {e}"
+            )));
+        }
+    }
+
```
aelesbao: You can move the `success()` call out of the match since the `Err` variant already `return` from the function.

```suggestion
        }
        Ok(false) => {
            // No config file yet; nothing to clear
        }
        Err(e) => {
            cleanup_handle.error();
            return Err(NexusCliError::Any(anyhow!(
                "Failed to check existence of CLI configuration: {e}"
            )));
        }
    }

    cleanup_handle.success();
```

CHANGELOG.md
```diff
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [`0.3.0`] - unreleased
```
aelesbao: Please do not add any version info the the changelog yet. This should happen at the release only. For example, we could decide to release `v0.2.1` at any time to fix bugs (including this one).

```suggestion
## Unreleased
```

davidrotari19: The pattern seems to be standard in all previous prs and in the new ones:
Example: 
- [2852711](https://github.com/Talus-Network/nexus/commit/28527110b53fc4995c2f44a11d9c7cebfdc3a4ca)
- [cde06d5](https://github.com/Talus-Network/nexus-sdk/commit/cde06d599f3d73bd7d6a3578073562bbc2b735d3)

loverdos: My mistake.

PR #290, opened 2025-09-01
feat: add onchain tools to nexus

nonast:
## Notable changes

- Added onchain tools support to nexus sdk
[see CHANGELOG.md for slightly more detailed changes]

## References

Fixes Epic https://github.com/Talus-Network/nexus/issues/339

## Checklist

- [x] keep a changelog
- [x] write tests

### DIFF
--- a/cli/src/tool/tool_register_onchain.rs
+++ b/cli/src/tool/tool_register_onchain.rs
@@ -1,7 +1,13 @@
 use {
     crate::{
-        command_title, display::json_output, loading, notify_error, notify_success, prelude::*,
-        sui::*, tool::tool_register_offchain,
+        command_title,
+        display::json_output,
+        loading,
+        notify_error,
+        notify_success,
+        prelude::*,
+        sui::*,
+        tool::tool_register_offchain,
     },
     nexus_sdk::{
         idents::{primitives, workflow},
--- a/sdk/src/sui.rs
+++ b/sdk/src/sui.rs
@@ -13,7 +13,11 @@ pub use {
         account_address::AccountAddress as MoveAccountAddress,
         ident_str as move_ident_str,
         identifier::{IdentStr as MoveIdentStr, Identifier as MoveIdentifier},
-        language_storage::{ModuleId as MoveModuleId, StructTag as MoveStructTag, TypeTag as MoveTypeTag},
+        language_storage::{
+            ModuleId as MoveModuleId,
+            StructTag as MoveStructTag,
+            TypeTag as MoveTypeTag,
+        },
         u256::U256 as MoveU256,
     },
     shared_crypto::intent::Intent,


cli/src/tool/mod.rs
```diff
@@ -11,12 +12,129 @@ use {
...
+            help = "If this flag is set, the tool owner caps will not be saved to the local config file."
+        )]
+        no_save: bool,
+
+        #[command(flatten)]
+        gas: GasArgs,
+    },
+
+    #[command(about = "Register an onchain tool")]
+    Onchain {
```
kouks, line 63: We probably want `collateral_coin` here too for onchain tools.

cli/src/tool/tool_register_offchain.rs
```diff
@@ -1,39 +1,27 @@
...
-        command_title,
-        display::json_output,
-        loading,
-        notify_error,
-        notify_success,
-        prelude::*,
-        sui::*,
-        tool::{tool_validate::*, ToolIdent},
+        command_title, display::json_output, loading, notify_error, notify_success, prelude::*,
+        sui::*, tool::tool_validate::validate_off_chain_tool,
```
kouks: `cargo +nightly fmt`

cli/src/tool/tool_register_onchain.rs
```diff
@@ -0,0 +1,1094 @@
...
+    let address = wallet.active_address().map_err(NexusCliError::Any)?;
+
+    // Fetch gas coin object.
+    let gas_coin = fetch_gas_coin(&sui, address, sui_gas_coin).await?;
+
+    // Fetch reference gas price.
+    let reference_gas_price = fetch_reference_gas_price(&sui).await?;
+
+    // Generate input schema by introspecting the Move module's "execute" function.
+    notify_success!("Auto-generating input schema from Move module...");
```
kouks: use `loading!` like in the other commands

cli/src/tool/tool_register_onchain.rs
```diff
@@ -0,0 +1,1094 @@
...
+                    );
+                    println!("{} Updated description", "✓".green().bold());
+                } else {
+                    println!("{} Kept current description", "→".truecolor(150, 150, 150));
+                }
+            }
+        }
+    }
+
+    println!(" ");
```
kouks:
```suggestion
    println!();
```

cli/src/tool/tool_register_onchain.rs
```diff
@@ -0,0 +1,1094 @@
...
+                                        "→".truecolor(150, 150, 150)
+                                    );
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
```
kouks, line 520: isn't there a better way to do this? feels a bet excessive. If not, would at least be nice to reduce the nesting, maybe extract a macro for the command line io interaction to read the updated values.

cli/src/tool/mod.rs
```diff
@@ -11,12 +12,129 @@ use {
...
+        #[arg(
+            long = "package-address",
+            short = 'p',
+            help = "The address of the published package containing the tool.",
+            value_name = "ADDRESS"
+        )]
```
davidrotari19: I would personally combine package and module name into a single generic address like in `url` for `off-chain` tool

cli/src/tool/tool_register_onchain.rs
kouks: we can reduce the nesting quite a bit by inverting logic checks. as in `if x.is_empty continue;`

cli/src/tool/tool_register_onchain.rs
```diff
@@ -0,0 +1,1094 @@
...
+
+    println!(" ");
```
kouks:
```suggestion
    println!();
```

cli/src/tool/templates/move/tool.move.jinja
```diff
@@ -0,0 +1,121 @@
...
+use sui::bag::{Self, Bag};
+use sui::transfer::share_object;
+use std::ascii::String as AsciiString;
+use sui::clock::Clock;
+
+/// One-time witness for package initialization.
+public struct {{ name_uppercase }} has drop {}
+
+/// Witness object used to identify this tool.
+public struct {{ name_pascal_case }}Witness has key, store {
```
davidrotari19, line 14: If you think about it, witness does not need to be an object. A simple drop struct will work just fine

cli/src/tool/tool_list.rs
```diff
@@ -75,15 +103,35 @@ pub(crate) async fn list_tools() -> AnyResult<(), NexusCliError> {
...
+struct OnChainTool {
+    package_address: String,
+    module_name: String,
+    witness_id: String,
+    #[serde(deserialize_with = "deserialize_bytes_to_lossy_utf8")]
+    description: String,
+    #[serde(deserialize_with = "deserialize_bytes_to_lossy_utf8")]
+    input_schema: String,
+    #[serde(deserialize_with = "deserialize_string_to_datetime")]
+    registered_at_ms: chrono::DateTime<chrono::Utc>,
```
davidrotari19, line 136: Should be `u64` as on-chain clock?

cli/src/tool/tool_register_onchain.rs
```diff
@@ -0,0 +1,1094 @@
...
+                fqn = fqn.to_string().truecolor(100, 100, 100),
+                error = e
+            );
+
+            return Err(e);
+        }
+    };
+
+    // Parse the owner cap object IDs from the response.
+    let owner_caps = response
```
davidrotari19, line 308: Would upgrade cap, also server as `tool cap`

sdk/src/idents/workflow.rs
```diff
@@ -468,6 +505,71 @@ impl Dag {
...
+        module: TOOL_OUTPUT_MODULE,
+        name: sui::move_ident_str!("new"),
+    };
+    /// Create an ok output variant.
+    ///
+    /// `nexus_workflow::tool_output::ok`
+    pub const OK: ModuleAndNameIdent = ModuleAndNameIdent {
+        module: TOOL_OUTPUT_MODULE,
+        name: sui::move_ident_str!("ok"),
+    };
```
kouks: why do we define these? can you exactly describe how the tool output works? because the names of these variants should ideally be definable by the tool owner

cli/src/tool/tool_register_onchain.rs
```diff
@@ -0,0 +1,1094 @@
...
+        )));
+    };
+
+    notify_success!(
+        "OwnerCap<OverTool> object ID: {id}",
+        id = over_tool_id.to_string().truecolor(100, 100, 100)
+    );
+
+    // Note: Onchain tools don't create OverGas caps, so we only save the OverTool cap
+    notify_success!("Onchain tools use a different gas model. No OverGas cap was created.");
```
davidrotari19, line 353: So `on-chain` tools are free?

sdk/src/object_crawler/fetching.rs
```diff
@@ -28,7 +28,7 @@ where
 }
 
 /// Fetch a dynamic field object from Sui based on the provided key.
-pub(crate) async fn dynamic_fetch_one<V>(
+pub async fn dynamic_fetch_one<V>(
```
kouks, line 31: why is this changed? it's an internal function of the object crawler and should ideally stay that way

sdk/tests/move/onchain_tool_test/sources/onchain_tool.move
```diff
@@ -0,0 +1,104 @@
+module onchain_tool::onchain_tool;
```
kouks, line 1: is this actually used for a test anywhere?

cli/src/tool/mod.rs
nonast: Added collateral to onchain tools: [fb664d0](https://github.com/Talus-Network/nexus-sdk/pull/290/commits/fb664d04b12dec4c509631b47729c9bd2378b412)

Necessary changes have been made in the nexus framework as well.

cli/src/tool/tool_register_offchain.rs
nonast: [30ae2e5](https://github.com/Talus-Network/nexus-sdk/pull/290/commits/30ae2e5986ec500900e55dd7ddac677e759184b1)

cli/src/tool/tool_register_onchain.rs
nonast: [dfb80dc](https://github.com/Talus-Network/nexus-sdk/pull/290/commits/dfb80dc44bbfbd8d770bc7428f147a8c981048f6)

cli/src/tool/tool_register_onchain.rs
nonast: [3c21d64](https://github.com/Talus-Network/nexus-sdk/pull/290/commits/3c21d6453997973fa7264aa6bd8f3a0b56d95428)

cli/src/tool/tool_register_onchain.rs
nonast: [3c21d64](https://github.com/Talus-Network/nexus-sdk/pull/290/commits/3c21d6453997973fa7264aa6bd8f3a0b56d95428)

cli/src/tool/tool_register_onchain.rs
nonast: Added some structs so we don't have to do manual json stuff and reduced nesting by extracting some helper functions. 
[8c00e0b](https://github.com/Talus-Network/nexus-sdk/pull/290/commits/8c00e0b9994f3fb53dce3a79ba7494d4e2c664b5)

cli/src/tool/mod.rs
nonast: I reused sui's `ModuleId` struct for this
[2ae823b](https://github.com/Talus-Network/nexus-sdk/pull/290/commits/2ae823b106dcabb0709d704889f2b1ac176cc0f1)

cli/src/tool/tool_list.rs
nonast: it is to show the date in a human readable way when you list all tools

sdk/src/idents/workflow.rs
nonast: Removed the unused constants, kept `TO_DAG_TYPES` as that was the only one used in the leader to compose the PTB. In the `ToolOutput` move module, we allow tool owners to create custom variants and fields. But two variants are there purely for convenience: `err()` and `ok()`. They are not necessary, and can be removed if you think we don't want to give tool devs any "hints" to what kind of output variants they could use. 
[7d0e04d](https://github.com/Talus-Network/nexus-sdk/pull/290/commits/7d0e04d485d4ae21cdffcf15bf54565a4a1fac60)

cli/src/tool/tool_register_onchain.rs
nonast: This is up for discussion. Currently, we don't have a separate gas model for onchain tools, because we put them in the same PTB as `SubmitOnChainToolEval` from the leader. Here, we take gas fees from the budget for the leader gas cost. Onchain tools also don't require any offchain computation, unlike offchain tools, so there is no cost to the tool dev. Anyone can technically execute onchain tools, so if someone doesn't want to pay for it (if we add a fee for onchain tools) they can just bypass nexus.

sdk/src/object_crawler/fetching.rs
nonast: Good catch, slipped through. Reverted it back. [ff03def](https://github.com/Talus-Network/nexus-sdk/pull/290/commits/ff03def7d693e9c79cc0159d1151ad99c82534e2)

sdk/tests/move/onchain_tool_test/sources/onchain_tool.move
nonast: Yes this is a slightly modified version of an onchain tool module used for tests in `input.rs` and `output.rs` to test the schema generation.

cli/src/tool/tool_register_onchain.rs
davidrotari19: >Anyone can technically execute onchain tools, so if someone doesn't want to pay for it (if we add a fee for onchain tools) they can just bypass nexus.

Wouldnt same happen to offchain, offchain is just a http server, so anybody can just call server directly

davidrotari19 (approved):
LGTM

cli/src/tool/tool_register_offchain.rs
loverdos: @nonast Reminder: the pre-commit checks should handle this. In fact it's what CI runs.

`just pre-commit` to see all the commands we run as part of the suite. Or `just pre-commit::cargo-nightly-fmt` for the specific one mentioned here.

@kouks @nonast Note that we do not do `cargo +nightly` anymore, we use a specific `nightly` snapshot/version. It's defined [here](https://github.com/Talus-Network/nexus-sdk/blob/ba888a7f1b04bc6a43d300d067cdac3066853112/.nightly-version) and it's used [here](https://github.com/Talus-Network/nexus-sdk/blob/ba888a7f1b04bc6a43d300d067cdac3066853112/.pre-commit/.just#L19-L21)

cli/src/tool/tool_register_onchain.rs
loverdos: Do we need a TODO somewhere?

kouks (approved):

PR #296, opened 2025-09-10
HTTP (any api) Tool

ebrartamer:
## Notable changes

- JSON schema validation and structured I/O support have been added.
- URL processing, authentication, and HTTP method defaults have been adjusted.
- Request body handling has been configured according to the HTTP method.
- Error handling and validation structure have been enhanced with structured enums.
- Retry logic and retry strategies have been added.

## Checklist

- [ ] ~~keep a changelog~~
- [x] write tests

### DIFF
--- a/tools/http/src/http.rs
+++ b/tools/http/src/http.rs
@@ -451,12 +451,16 @@ impl Http {
 mod tests {
     use {super::*, mockito::Server};
 
+    /// Helper function to create a mock server and HTTP tool for testing
+    async fn create_server_and_tool() -> (mockito::ServerGuard, Http) {
+        let server = Server::new_async().await;
+        let tool = Http::new().await;
+        (server, tool)
+    }
+
     #[tokio::test]
     async fn test_http_get() {
-        let tool = Http::new().await;
-
-        // Create mock server
-        let mut server = Server::new_async().await;
+        let (mut server, tool) = create_server_and_tool().await;
         let mock_response = r#"{"method": "GET", "url": "http://example.com/get", "args": {}}"#;
         let _mock = server
             .mock("GET", "/get")
@@ -509,10 +513,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_default_http_method() {
-        let tool = Http::new().await;
-
-        // Create mock server
-        let mut server = Server::new_async().await;
+        let (mut server, tool) = create_server_and_tool().await;
         let mock_response = r#"{"method": "GET", "url": "http://example.com/default", "args": {}}"#;
         let _mock = server
             .mock("GET", "/default")
@@ -566,10 +567,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_http_head() {
-        let tool = Http::new().await;
-
-        // Create mock server - HEAD requests typically don't return body
-        let mut server = Server::new_async().await;
+        let (mut server, tool) = create_server_and_tool().await;
         let _mock = server
             .mock("HEAD", "/head")
             .with_status(200)
@@ -619,10 +617,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_http_404_error() {
-        let tool = Http::new().await;
-
-        // Create mock server for 404 error
-        let mut server = Server::new_async().await;
+        let (mut server, tool) = create_server_and_tool().await;
         let _mock = server
             .mock("GET", "/notfound")
             .with_status(404)
@@ -688,10 +683,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_json_parse_error() {
-        let tool = Http::new().await;
-
-        // Create mock server that returns invalid JSON
-        let mut server = Server::new_async().await;
+        let (mut server, tool) = create_server_and_tool().await;
         let _mock = server
             .mock("GET", "/invalid-json")
             .with_status(200)
@@ -726,10 +718,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_url_split() {
-        let tool = Http::new().await;
-
-        // Create mock server
-        let mut server = Server::new_async().await;
+        let (mut server, tool) = create_server_and_tool().await;
         let mock_response =
             r#"{"method": "GET", "url": "http://example.com/api/users", "args": {}}"#;
         let _mock = server
@@ -781,10 +770,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_headers_and_query() {
-        let tool = Http::new().await;
-
-        // Create mock server
-        let mut server = Server::new_async().await;
+        let (mut server, tool) = create_server_and_tool().await;
         let mock_response = r#"{"method": "GET", "url": "http://example.com/api/users?page=1&limit=10", "headers": {"Authorization": "Bearer token123", "Content-Type": "application/json"}}"#;
         let _mock = server
             .mock("GET", "/api/users")
@@ -846,10 +832,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_auth_bearer_token() {
-        let tool = Http::new().await;
-
-        // Create mock server
-        let mut server = Server::new_async().await;
+        let (mut server, tool) = create_server_and_tool().await;
         let mock_response = r#"{"authenticated": true, "token": "test-token"}"#;
         let _mock = server
             .mock("GET", "/auth")
@@ -887,10 +870,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_auth_api_key_header() {
-        let tool = Http::new().await;
-
-        // Create mock server
-        let mut server = Server::new_async().await;
+        let (mut server, tool) = create_server_and_tool().await;
         let mock_response = r#"{"authenticated": true, "api_key": "test-key"}"#;
         let _mock = server
             .mock("GET", "/auth")
@@ -929,10 +909,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_auth_api_key_query() {
-        let tool = Http::new().await;
-
-        // Create mock server
-        let mut server = Server::new_async().await;
+        let (mut server, tool) = create_server_and_tool().await;
         let mock_response = r#"{"authenticated": true, "api_key": "test-key"}"#;
         let _mock = server
             .mock("GET", "/auth")
@@ -972,10 +949,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_auth_basic() {
-        let tool = Http::new().await;
-
-        // Create mock server
-        let mut server = Server::new_async().await;
+        let (mut server, tool) = create_server_and_tool().await;
         let mock_response = r#"{"authenticated": true, "user": "testuser"}"#;
         let _mock = server
             .mock("GET", "/auth")
@@ -1131,10 +1105,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_json_body() {
-        let tool = Http::new().await;
-
-        // Create mock server
-        let mut server = Server::new_async().await;
+        let (mut server, tool) = create_server_and_tool().await;
         let mock_response = r#"{"method": "POST", "url": "http://example.com/post", "data": {"name": "test", "value": 123}}"#;
         let _mock = server
             .mock("POST", "/post")
@@ -1174,10 +1145,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_raw_body() {
-        let tool = Http::new().await;
-
-        // Create mock server
-        let mut server = Server::new_async().await;
+        let (mut server, tool) = create_server_and_tool().await;
         let mock_response =
             r#"{"method": "POST", "url": "http://example.com/post", "data": "binary data"}"#;
         let _mock = server
@@ -1282,10 +1250,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_timeout_configuration() {
-        let tool = Http::new().await;
-
-        // Create mock server
-        let mut server = Server::new_async().await;
+        let (mut server, tool) = create_server_and_tool().await;
         let mock_response = r#"{"method": "GET", "url": "http://example.com/get", "args": {}}"#;
         let _mock = server
             .mock("GET", "/get")
@@ -1321,10 +1286,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_retries_configuration() {
-        let tool = Http::new().await;
-
-        // Create mock server
-        let mut server = Server::new_async().await;
+        let (mut server, tool) = create_server_and_tool().await;
         let mock_response = r#"{"method": "GET", "url": "http://example.com/get", "args": {}}"#;
         let _mock = server
             .mock("GET", "/get")
@@ -1360,10 +1322,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_retry_on_server_errors() {
-        let tool = Http::new().await;
-
-        // Create mock server that returns 500 error first, then 200
-        let mut server = Server::new_async().await;
+        let (mut server, tool) = create_server_and_tool().await;
         let mock_response = r#"{"method": "GET", "url": "http://example.com/get", "args": {}}"#;
 
         // First request returns 500
@@ -1411,10 +1370,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_no_retry_on_client_errors() {
-        let tool = Http::new().await;
-
-        // Create mock server that returns 404 error
-        let mut server = Server::new_async().await;
+        let (mut server, tool) = create_server_and_tool().await;
         let _mock = server
             .mock("GET", "/notfound")
             .with_status(404)
@@ -1456,10 +1412,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_get_method_ignores_body() {
-        let tool = Http::new().await;
-
-        // Create mock server
-        let mut server = Server::new_async().await;
+        let (mut server, tool) = create_server_and_tool().await;
         let mock_response = r#"{"method": "GET", "url": "http://example.com/get", "args": {}}"#;
         let _mock = server
             .mock("GET", "/get")
@@ -1501,10 +1454,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_head_method_ignores_body() {
-        let tool = Http::new().await;
-
-        // Create mock server
-        let mut server = Server::new_async().await;
+        let (mut server, tool) = create_server_and_tool().await;
         let _mock = server
             .mock("HEAD", "/head")
             .with_status(200)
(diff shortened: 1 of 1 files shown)


tools/http/README.md
```diff
@@ -0,0 +1,155 @@
...
+- **`Json { data }`** - JSON request body
+- **`Form { data }`** - URL-encoded form data
+- **`Multipart { fields }`** - Multipart form data for file uploads
+- **`Raw { data, content_type }`** - Raw bytes (base64 encoded)
+
+### MultipartField Structure
+
+- **`name`: [`String`]** - Field name
+- **`value`: [`String`]** - Field value (text) or base64 encoded data (for files)
+- **`filename`: [`Option<String>`]** - Filename for file uploads
```
aelesbao: I don't think we should support file uploads at this moment.

tools/http/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+Request timeout in milliseconds.
+
+_opt_ **`retries`: [`Option<u32>`]** _default_: [`0`]
+
+Number of retries on failure.
+
+_opt_ **`follow_redirects`: [`Option<bool>`]** _default_: [`true`]
+
+Whether to follow HTTP redirects.
```
aelesbao: `curl` does not follow HTTP redirects by default. This is based on its tradition of only doing the basics unless you tell it differently. IMHO, we should follow curl's philosophy in this tool design.

tools/http/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+### HttpErrorKind Values
+
+- **`err_http`** - HTTP error response (4xx, 5xx)
+- **`err_json_parse`** - JSON parsing error
+- **`err_schema_validation`** - Schema validation error
+- **`err_network`** - Network connectivity error
+- **`err_input`** - Input validation error
+- **`err_url_parse`** - URL parsing error
+- **`err_base64_decode`** - Base64 decoding error
```
aelesbao, line 155: We should have an error type for timeouts

tools/http/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+Number of retries on failure.
+
+_opt_ **`follow_redirects`: [`Option<bool>`]** _default_: [`true`]
+
+Whether to follow HTTP redirects.
+
+_opt_ **`allow_empty_json`: [`Option<bool>`]** _default_: [`false`]
+
+If true, tolerate empty body when JSON is expected (e.g., 204 No Content).
```
aelesbao: Do you have a practical example of why this flag is needed?

tools/http/src/http.rs
```diff
@@ -0,0 +1,1674 @@
...
+
+    fn description() -> &'static str {
+        "Generic HTTP tool that can make requests to any API endpoint."
+    }
+
+    async fn health(&self) -> AnyResult<StatusCode> {
+        Ok(StatusCode::OK)
+    }
+
+    async fn invoke(&self, input: Self::Input) -> Self::Output {
```
aelesbao, line 219: Please break down this function into [smaller functions](https://ashishmd.medium.com/summary-of-clean-code-by-robert-c-martin-part-3-functions-72b3115472e) for better readability (it has almost 200 LOC).

tools/http/src/http_client.rs
```diff
@@ -0,0 +1,261 @@
...
+    /// Creates a new HTTP client with custom configuration
+    pub fn with_config(
+        timeout_ms: Option<u64>,
+        follow_redirects: Option<bool>,
+    ) -> Result<Self, HttpToolError> {
+        let mut builder = Client::builder();
+
+        // Set timeout with default (30 seconds = 30000ms)
+        let timeout_ms = timeout_ms.unwrap_or(30000);
+        builder = builder.timeout(std::time::Duration::from_millis(timeout_ms));
```
aelesbao, line 38: We should see an upper boundary for the timeout. For the purpose of this tool, 30 sec should be the maximum, and we can reduce the default value to 5 seconds.

tools/http/src/http_client.rs
```diff
@@ -0,0 +1,261 @@
...
+                            let part = reqwest::multipart::Part::text(field.value.clone());
+                            form = form.part(field.name.clone(), part);
+                        }
+                    } else {
+                        // Text field without filename
+                        let part = reqwest::multipart::Part::text(field.value.clone());
+                        form = form.part(field.name.clone(), part);
+                    }
+                }
+                Ok(request.multipart(form))
```
aelesbao: Please break this into a separate function

tools/http/src/utils.rs
```diff
@@ -0,0 +1,64 @@
...
+    serde_json::Value,
+};
+
+/// Validate JSON data against a schema and return detailed results
+pub(crate) fn validate_schema_detailed(
+    schema_def: &HttpJsonSchema,
+    json_data: &Value,
+) -> SchemaValidationDetails {
+    // Convert schema to JSON value for validation
+    let schema_value = match serde_json::to_value(&schema_def.schema) {
```
aelesbao: Instead of using `match`, you can write a shorter version with [`map_err`](https://doc.rust-lang.org/std/result/enum.Result.html#method.map_err):

```rust
serde_json::to_value(&schema_def.schema)
  .map_err(e => SchemaValidationDetails {
                name: schema_def.name.clone(),
                description: schema_def.description.clone(),
                strict: schema_def.strict,
                valid: false,
                errors: vec!["Schema serialization failed".to_string()],
  })?;
```

tools/http/src/utils.rs
```diff
@@ -0,0 +1,64 @@
...
+                description: schema_def.description.clone(),
+                strict: schema_def.strict,
+                valid: false,
+                errors: vec!["Schema serialization failed".to_string()],
+            };
+        }
+    };
+
+    // Validate using jsonschema
+    let validator = match jsonschema::validator_for(&schema_value) {
```
aelesbao: Same as [here](https://github.com/Talus-Network/nexus-sdk/pull/296/files#r2376318942)

tools/http/src/http_client.rs
```diff
@@ -0,0 +1,261 @@
...
+            HttpToolError::ErrInput("Request failed after all retries".to_string())
+        }))
+    }
+}
+
+impl Default for HttpClient {
+    fn default() -> Self {
+        Self::new().expect("Failed to create HTTP client")
+    }
+}
```
aelesbao, line 261: Please add some tests for the `HttpClient`

tools/http/src/http_client.rs
```diff
@@ -0,0 +1,261 @@
...
+        let mut builder = Client::builder();
+
+        // Set timeout with default (30 seconds = 30000ms)
+        let timeout_ms = timeout_ms.unwrap_or(30000);
+        builder = builder.timeout(std::time::Duration::from_millis(timeout_ms));
+
+        // Set redirect policy with default (follow redirects)
+        let follow_redirects = follow_redirects.unwrap_or(true);
+        if follow_redirects {
+            builder = builder.redirect(reqwest::redirect::Policy::limited(10));
```
aelesbao: We should allow a smaller number of redirects by default, probably 3. Every decent API will not have more than 1 redirect.

Also, is the timeout applied to the entire request or each of the redirects will have their own timeout?

tools/http/src/http.rs
```diff
@@ -0,0 +1,1674 @@
...
+                Ok(request) => request,
+                Err(e) => return e.to_output(),
+            }
+        } else {
+            request
+        };
+
+        // Execute request with or without retry logic
+        let retries = input.retries.unwrap_or(0);
+        let response = if retries > 0 {
```
aelesbao: We should not allow more than 5 retries.

tools/http/src/http.rs
```diff
@@ -0,0 +1,1674 @@
...
+            http_client.execute_with_retry(request, retries).await
+        } else {
+            http_client.execute(request).await
+        };
+
+        match response {
+            Ok(response) => {
+                let status = response.status().as_u16();
```
aelesbao: Please use the [`StatusCode`](https://docs.rs/reqwest/latest/reqwest/struct.StatusCode.html) constants instead of numeric codes.

tools/http/README.md
berk-turan: We simplified the multipart feature to support only text-based form data, removing the relevant fields from the MultipartField struct and eliminating base64 decoding logic from the HttpClient.

fixed. 1d70153

tools/http/README.md
berk-turan: Updated the default value of follow_redirects from true to false in both the HttpClient and main tool logic, to align with curl’s philosophy of not following redirects by default. Modified tests to explicitly specify follow_redirects: true where needed.

fixed. 64c8752776892da985bc4c1e0232abf08d6f61ee

tools/http/README.md
berk-turan: Added a dedicated timeout error type by introducing a Timeout variant in the HttpErrorKind enum and an ErrTimeout variant in the HttpToolError enum. Enhanced timeout detection logic using both the e.is_timeout() method and error message analysis.

fixed. f9887fb4a5e19cddb089045547fc9c312360c288

tools/http/README.md
berk-turan: We think that this flag is necessary because some real-world APIs return **`204 No Content`** with an empty body even when the client expects JSON. Without the flag, the tool would attempt to parse the empty response and fail.

**Examples where 204 may be returned:**
- **DELETE operations:**  
  - GitHub: Deleting a repository via `DELETE /repos/{owner}/{repo}` → [docs](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#delete-a-repository)  
  - Twitter: Deleting a direct message event via `DELETE /1.1/direct_messages/events/destroy.json` → [docs](https://developer.x.com/en/docs/x-api/v1/direct-messages/sending-and-receiving/api-reference/delete-message-event)  
- **PUT operations:**  
  - Jira: Updating an issue via `PUT /rest/api/2/issue/{id}` → [docs](https://developer.atlassian.com/server/jira/platform/jira-rest-api-example-edit-issues-6291632/)  
  - Some other update APIs return 204 when successful  
- **PATCH operations:** Partial updates that return 204 when successful  
- **POST operations:** Webhook acknowledgments, health checks, or some create endpoints  

The `allow_empty_json` flag ensures these cases work smoothly when `expect_json: true` is set.  

Does this reasoning make sense, or would you suggest a different approach? We’d love to hear your thoughts.

tools/http/src/http.rs
berk-turan: Refactored the function into 4 smaller, single-responsibility functions: 

- `prepare_request` for HTTP client setup and request building, 
- `process_response` for response processing and headers, 
- `parse_json_response` for JSON parsing logic, 
- `validate_schema_response` for schema validation. 

The main `invoke` function is now reduced to 25 lines. 

I think it is fixed with e378653. What do you think? @afarukcali

tools/http/src/http_client.rs
berk-turan: Reduced default timeout from 30s to 5s, and limited maximum timeout from 5 minutes to 30s for better resource management. Added proper validation to enforce the 30s upper limit.

fixed. ecf160558f5bd9bbdb36e0d7c3cf8be53ebe86e9

tools/http/src/http_client.rs
berk-turan: Refactored HTTP client by extracting multipart form building and raw body handling into separate functions (`build_multipart_form()` and `build_raw_body()`)

fixed. b418c5f842199deeaf7260b3506701e16b510a2e

tools/http/src/http_client.rs
berk-turan: Added tests for the HttpClient. They cover client creation, URL resolution, HTTP method handling, request building with authentication/headers/query parameters, and all body types (JSON, form, multipart, raw).

### Added Tests:
- `test_http_client_new()` - Basic client creation
- `test_http_client_with_config()` - Custom configuration
- `test_resolve_url_full()` - Full URL parsing
- `test_resolve_url_split()` - Split URL parsing
- `test_resolve_url_invalid()` - Invalid URL handling
- `test_build_method()` - All HTTP methods
- `test_method_supports_body()` - Body support validation
- `test_build_request_with_auth()` - Authentication handling
- `test_build_request_with_headers()` - Custom headers
- `test_build_request_with_query()` - Query parameters
- `test_build_multipart_form()` - Multipart form handling
- `test_build_raw_body()` - Raw body (base64) handling
- `test_build_raw_body_invalid_base64()` - Error handling
- `test_build_body_json()` - JSON body handling
- `test_build_body_form()` - Form body handling
- `test_build_body_ignores_for_get()` - GET method body handling

fixed. de84ea2653fbd040a59feaa9f527819be61cd8c3

tools/http/src/http.rs
berk-turan: Replaced numeric status code comparisons with reqwest's StatusCode enum methods for better type safety and readability.

fixed. 62f8e1c60f2741d0514edefcb0c483c6ddf53086

tools/http/src/http.rs
berk-turan: Updated retry validation to enforce a maximum of 5 retries as per feedback. Changed the validation logic from allowing up to 10 retries to a maximum of 5 retries, improving resource management and preventing excessive retry attempts.

fixed. 878faafac11335f55f679ea2fe60304f2b61a124

tools/http/src/utils.rs
berk-turan: Refactored verbose `match` statements to use `map_err` for cleaner error handling.  
Key changes:  
- Updated `validate_schema_detailed` return type to `Result<SchemaValidationDetails, SchemaValidationDetails>`  
- Replaced `match` blocks with `map_err` calls  
- Added `From<SchemaValidationDetails>` implementation for `HttpToolError`  

fixed. 252f6bc5bd5ad1b5e90174620e60caa4c39e0425

tools/http/src/utils.rs
berk-turan: fixed. 252f6bc5bd5ad1b5e90174620e60caa4c39e0425

tools/http/src/http_client.rs
berk-turan: We changed the default limit from 10 to 3. The `timeout_ms` parameter applies to the **entire request chain**, not to individual redirects. When `follow_redirects` is enabled, the timeout covers the complete flow — from the initial request through all redirects until the final response.  

For example: if `timeout_ms: 5000` is set and a request follows 3 redirects (each taking ~1 second), the total time limit remains **5 seconds for the whole chain**, not 5 seconds per redirect. This behavior prevents infinite redirect loops and ensures predictable response times. The timeout is enforced at the `reqwest` client level using:  

```rust
builder.timeout(std::time::Duration::from_millis(timeout_ms))
```
and applies to the entire HTTP transaction. You can find relevant documents here.

**Reference:** [reqwest::Client::timeout](https://docs.rs/reqwest/latest/reqwest/struct.Client.html#method.timeout) - "Sets a timeout for the entire request."

What is your opinion. Should there anything else to be changed?

fixed. 6451be8d4d1668778da4d0907b05aeee9b966e75

berk-turan:
> > My main concern with this generic HTTP tool is that it creates an attack surface by not blocking workflow users to run requests on internal network endpoints (like the Kubernetes or Docker APIs, or even localhost). We should add some safe-guard to block those kind of malicious requests in the code by identifying when the base URL points to an internal IP.
> 
> @berk-turan I can help you with researching this topic. Once the small comments are completed, let’s discuss this matter between us.

We completed requested changes. May you review it? If we need to change some parts of them, please let us know. 

After your review, we can discuss about the main concern. We need your opinion to research well.

tools/http/src/http.rs
afarukcali: Thank you @berk-turan . It looks good now, but @aelesbao 's feedback is important.

tools/http/README.md
aelesbao: In this case, we can remove the flag and just look at the response status code. The status [204 literally means "No Content"](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/204). Please pay attention that there are other successful status codes that also don't return content, like [202](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/202) or [205](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/205).

tools/http/src/http.rs
```diff
@@ -0,0 +1,1768 @@
...
+                assert!(reason.contains("Failed to parse JSON"));
+            }
+            Output::Ok { text, .. } => {
+                // If response is not JSON but text, we should not get a JSON parse error
+                if let Some(text_content) = text {
+                    if !text_content.trim().is_empty() {
+                        // Text response but not JSON - this is normal
+                    }
+                }
+            }
```
aelesbao: What's the purpose of this empty matcher?

tools/http/src/http_client.rs
```diff
@@ -0,0 +1,500 @@
...
+        retries: u32,
+    ) -> Result<reqwest::Response, HttpToolError> {
+        if retries == 0 {
+            // No retries needed, execute once
+            return self.execute(request).await;
+        }
+
+        let mut last_error = None;
+
+        for attempt in 0..=retries {
```
aelesbao: Instead of writing our own retry logic, let's simplify the code using [`backon`](https://docs.rs/backon/latest/backon/) for it.

aelesbao (changes requested):

tools/http/src/http.rs
berk-turan: It was entirely my mistake; I left it there while working on the tests. Thanks for catching it.

fixed. 73acbda8a6d685d3bacf7261d931b33046e1c25c

tools/http/README.md
berk-turan: We removed the flag, we are just looking at the status code now. 

fixed. 435431b21fc7142e47c448970139f022633fcbf1

tools/http/src/http_client.rs
berk-turan: Since it wasn’t in the main workspace, we didn’t want to add a new dependency, so we updated the retry logic using backon.

fixed. 5129aba8e493cebc3467327fa9087172df0e48ab

tools/http/src/http_client.rs
```diff
@@ -0,0 +1,489 @@
...
+#[cfg(test)]
+mod tests {
+    use {
+        super::*,
+        crate::models::{AuthConfig, HttpMethod, RequestBody, UrlInput},
+        std::collections::HashMap,
+    };
+
+    #[test]
+    fn test_http_client_new() {
```
afarukcali, line 272: @berk-turan  i think, we can use `mockito` for unit tests. Isn’t that right @aelesbao ?

afarukcali (changes requested):

tools/http/src/http_client.rs
aelesbao: Sure, `mockito` is in the workspace crates, feel free to use it

berk-turan: We fixed unrelated tests which uses httpbin to make requests to mock server. We use mockito for all tests. 

fixed. 2ece3bfa732fa2e2d2d063659a19ba24b2e1e6eb

aelesbao (approved):

berk-turan:
We added a helper function to separate creating mock server and tool.

b99c3ed8d7604793d62d30b079a6677c5e2b06dc

afarukcali (approved):

PR #302, opened 2025-10-09
[0.3] Remote storage support storage in SDK and CLI

kouks:
## Notable changes

- see `CHANGELOG.md`

## References

- resolves https://github.com/Talus-Network/nexus/issues/516
- resolves https://github.com/Talus-Network/nexus/issues/518
- resolves https://github.com/Talus-Network/nexus/issues/517
- resolves https://github.com/Talus-Network/nexus/issues/520
- resolves https://github.com/Talus-Network/nexus/issues/519

## Checklist

- [x] keep a changelog
- [x] write tests

### DIFF
--- a/sdk/tests/walrus_client_test.rs
+++ b/sdk/tests/walrus_client_test.rs
@@ -15,7 +15,7 @@ struct SampleData {
     value: i32,
 }
 
-const EPOCHS: u64 = 1;
+const EPOCHS: u8 = 1;
 const TEST_CONTENT: &[u8] = b"Hello, World!";
 
 /// Setup mock server for Walrus testing
--- a/tools/storage-walrus/src/upload_file.rs
+++ b/tools/storage-walrus/src/upload_file.rs
@@ -48,13 +48,13 @@ pub(crate) struct Input {
     publisher_url: Option<String>,
     /// The number of epochs to store the file
     #[serde(default = "default_epochs")]
-    epochs: u64,
+    epochs: u8,
     /// Optional address to which the created Blob object should be sent
     #[serde(default)]
     send_to: Option<String>,
 }
 
-fn default_epochs() -> u64 {
+fn default_epochs() -> u8 {
     1
 }
 
--- a/tools/storage-walrus/src/upload_json.rs
+++ b/tools/storage-walrus/src/upload_json.rs
@@ -43,13 +43,13 @@ pub(crate) struct Input {
     aggregator_url: Option<String>,
     /// Number of epochs to store the data
     #[serde(default = "default_epochs")]
-    epochs: u64,
+    epochs: u8,
     /// Optional address to which the created Blob object should be sent
     #[serde(default)]
     send_to_address: Option<String>,
 }
 
-fn default_epochs() -> u64 {
+fn default_epochs() -> u8 {
     1
 }
 


cli/src/conf/conf_set.rs
```diff
@@ -1,11 +1,22 @@
...
 
 /// Set the Nexus CLI configuration from the provided arguments.
 pub(crate) async fn set_nexus_conf(
     sui_net: Option<SuiNet>,
     sui_wallet_path: Option<PathBuf>,
     sui_rpc_url: Option<reqwest::Url>,
     nexus_objects_path: Option<PathBuf>,
+    data_storage_walrus_aggregator_url: Option<reqwest::Url>,
+    data_storage_walrus_publisher_url: Option<reqwest::Url>,
+    data_storage_walrus_save_for_epochs: Option<u64>,
```
davidrotari19: Its meant per epochs?

sdk/src/types/nexus_data.rs
```diff
@@ -2,27 +2,485 @@
...
 //!
-//! See: <https://github.com/Talus-Network/nexus-next/issues/30>
+//! The [`DataStorage`] enum has multiple implementations to support inline
+//! or remote storage.
 //!
-//! The [`NexusData::data`] field is a byte array on-chain but we assume that,
-//! upon decoding it, it will be a valid JSON object.
+//! Note: As an optimization to reduce storage costs, array types are stored as
+//! one blob containing a JSON array, instead of multiple blobs. This means that
+//! the array of keys referencing the data will contain only one key repeated N
```
davidrotari19, line 10: Should we do like one key and indexing? I guess that would require a bit of change so maybe for later.

cli/src/conf/conf_set.rs
kouks: it's the storage period so I believe `for` is correct?

sdk/src/types/nexus_data.rs
kouks: You might have to explain a bit more - one key for what?

cli/src/conf/conf_set.rs
davidrotari19: I thought it meant saves per epoch.

sdk/src/types/nexus_data.rs
davidrotari19: It was me being confused of how its saved. I looked at code and make sense. I was confused of why would you have a vector of keys if you save the whole array as a blob.

davidrotari19 (approved):
LGTM

sdk/src/types/nexus_data.rs
kouks: It's the same key cloned N times for keep the information about the # of iterations for the on-chain workflow

davidrotari19: I realized you use it as an iterator which fits nicely with everything else.

cli/src/prelude.rs
```diff
@@ -107,35 +114,96 @@ impl Default for SuiConf {
...
+        tokio::fs::create_dir_all(parent_folder).await?;
+        tokio::fs::write(path, conf).await?;
+
+        Ok(())
     }
 }
 
-impl Eq for CryptoConf {}
+/// Gets the active session for encryption/decryption
+pub async fn get_active_session() -> Result<Arc<Mutex<Session>>, NexusCliError> {
```
davidrotari19: Looks good. I was actually think of just taking it out and keeping the Identity in, works the same principle as Mutex but less redefined. So not the best idea

kouks: I have some more changes actually because I needed a good way to re-save the session

kouks: https://github.com/Talus-Network/nexus-sdk/pull/302/commits/8b8ddb43fc7afab84d42005a72c30c76cd44dec6

Cargo.toml
```diff
@@ -9,7 +9,7 @@ members = [
 
 [workspace.package]
 edition = "2021"
-version = "0.3.0-rc.1"
+version = "0.3.0-rc.2"
```
aelesbao: no need to bump the version since we don't have a tag yet

CHANGELOG.md
```diff
@@ -11,18 +11,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+#### Fixed
+
+- `nexus crypto init-key --force` wipes the old `crypto` state from config before rotating the key to avoid parsing errors
+
+### `nexus-sdk`
 
 #### Added
 
 - .nightly-version
```
aelesbao: I know this is not part of this PR, but it would be helpful to add more context to this changelog. cc @loverdos

sdk/src/types/nexus_data.rs
```diff
@@ -2,27 +2,553 @@
...
+    ///
+    /// let data = NexusData::new_inline(serde_json::json!({"key": "value"})).commit_inline_plain();
+    ///
+    /// assert_eq!(data.as_json(), &serde_json::json!({"key": "value"}));
+    /// assert!(!data.is_encrypted());
+    /// assert_eq!(data.storage_kind(), StorageKind::Inline);
+    /// ```
+    pub fn commit_inline_plain(self) -> DataStorage {
+        if self.data.storage_kind() != StorageKind::Inline || self.data.is_encrypted() {
+            panic!("commit_inline_plain can only be used for inline, unencrypted data");
```
aelesbao, line 94: We should avoid panicking in the SDK code. Let's use the 'Result' enum instead.

sdk/src/types/nexus_data.rs
```diff
@@ -2,27 +2,553 @@
...
+            "Cannot fit data within max transaction size, even after storing all fields remotely"
+        );
+    }
+
+    Ok(remote_fields)
+}
+
+// == Parsing ==
+
 mod parser {
```
aelesbao: This file has more than 1k LOC. Since this is a clear sub-module (even with its own sub-tests), this could be a great place to split the code into a new file.

cli/src/cli_conf.rs
```diff
@@ -0,0 +1,530 @@
...
+        std::env::set_var("XDG_CONFIG_HOME", secret_home.path());
+        std::env::set_var("XDG_DATA_HOME", secret_home.path());
+        secret_home
+    }
+
+    fn cleanup_env() {
+        std::env::remove_var("NEXUS_CLI_STORE_PASSPHRASE");
+        std::env::remove_var("XDG_CONFIG_HOME");
+        std::env::remove_var("XDG_DATA_HOME");
+    }
```
aelesbao, line 246: We should not temper the environment for testing purposes. Let's use dependency injection for it instead.

aelesbao (changes requested):

sdk/src/types/nexus_data.rs
kouks: If you check the doc comment, the entire purpose of the function is to be synchronous and non-result. I needed it for a specific case in the leader. Since it's intended and properly documented I'd like to keep it.

cli/src/cli_conf.rs
kouks: I copied the approach from David's crypto tests. @davidrotari19 do you have more insight here? I don't see a good way to inject the keystore.

sdk/src/types/nexus_data.rs
kouks: https://github.com/Talus-Network/nexus-sdk/pull/302/commits/946520b3526bd997d6ae3c710b410f7f24f8af37

CHANGELOG.md
kouks: https://github.com/Talus-Network/nexus-sdk/pull/302/commits/946520b3526bd997d6ae3c710b410f7f24f8af37

Cargo.toml
kouks: https://github.com/Talus-Network/nexus-sdk/pull/302/commits/946520b3526bd997d6ae3c710b410f7f24f8af37

cli/src/cli_conf.rs
davidrotari19: Dependency injection would require a lot of change and give no benefit in my opinion.

kouks: Tests run in serial based on the `master_env` key so there should be no clash

aelesbao: @kouks, that's fine, let's move forward without this refactoring for now.

@davidrotari19, Modifying the env makes the tests interdependent and prevents them from running safely in parallel, which is still an issue in `nexus`. Applying DI ensures that tests are isolated, deterministic, and side-effect-free, which should be our default approach going forward.

aelesbao (approved):

cli/src/cli_conf.rs
davidrotari19: >@davidrotari19, Modifying the env makes the tests interdependent and prevents them from running safely in parallel, which is still an issue in nexus. Applying DI ensures that tests are isolated, deterministic, and side-effect-free, which should be our default approach going forward.

Its true but this is not the only way to achieve this. You can namespace the test, that type of isolation is much easier to achieve and doesn't require dependency injection.

loverdos: My 2c. 

Env vars are a convenience to get to a configuration object. This functionality can be tested once, to make sure the configuration is read as intended. But all the tests should rely on the configuration _object_, so _after_ the env has been read into a Rust structure. 

Tweaking the env vars is not something I would recommend, they have well-known problems as global state that cannot be easily controlled in a concurrent environment.

So we need a ticket to refactor this whole approach.

@aelesbao @davidrotari19 @kouks

davidrotari19: https://github.com/Talus-Network/nexus-sdk/issues/313

PR #312, opened 2025-10-27
[0.3] Add `NexusClient`

kouks:
## Notable changes

- see `CHANGELOG.md`

## References

- resolves https://github.com/Talus-Network/nexus/issues/548
- resolves https://github.com/Talus-Network/nexus/issues/545

## Checklist

- [x] keep a changelog
- [x] write tests

### DIFF
--- a/sdk/tests/object_crawler_test.rs
+++ b/sdk/tests/object_crawler_test.rs
@@ -72,7 +72,7 @@ enum HeterogeneousValue {
     AnotherValue(AnotherPlainValue),
 }
 
-// TODO: <https://github.com/Talus-Network/nexus/issues/559>
+// TODO: <https://github.com/Talus-Network/nexus-sdk/issues/318>
 #[tokio::test]
 async fn test_object_crawler() {
     // Spin up the Sui instance.


sdk/src/nexus/client.rs
```diff
@@ -0,0 +1,575 @@
...
+    }
+
+    /// Release a gas coin back to the pool.
+    pub(super) async fn release_gas_coin(&self, coin: sui::ObjectRef) {
+        self.coins.lock().await.push(coin);
+        self.notify.notify_one();
+    }
+
+    /// Get the gas budget.
+    pub(super) fn get_budget(&self) -> u64 {
```
davidrotari19, line 179: The budget should be updated after release?

kouks: I was just thinking this would be a budget per-transaction so unchanging

davidrotari19: I see, you meant the gas budget passed in.

cli/src/dag/dag_execute.rs
```diff
@@ -208,33 +165,18 @@ async fn process_entry_ports(
...
-                    command = "$ nexus conf set --data-storage.walrus-publisher-url <URL> --data-storage.walrus-save-for-epochs <EPOCHS>",
-                    testnet_command = "Or for testnet simply: $ nexus conf set --data-storage.testnet"
-                ))
-            })?;
-
-        result.insert(vertex.clone(), ports_data);
+        result.insert(vertex.clone(), PortsData::from_map(nexus_data_map));
     }
 
     // Hint the user if they should use remote storage and for what fields.
```
davidrotari19, line 171: This comment seems wrong?

sdk/src/events.rs
```diff
@@ -77,6 +77,19 @@ pub enum NexusEventKind {
     AllowedOwnerRemoved(serde_json::Value),
 }
 
+impl NexusEventKind {
```
davidrotari19, line 80: Why not `Display`?

cli/src/dag/dag_execute.rs
kouks: Does it? The function call is underneath this code block that flattens the input data but that is exactly what is happening here.

sdk/src/events.rs
kouks: To me it doesn't semantically feel like a `impl Display` since we're ignoring all of the inner data and only returning the kind name

davidrotari19: Makes sense

PR #332, opened 2025-12-02
grpc: transaction templates

kouks:
## Notable changes

- adjust transaction templates to use the new SDK

### DIFF
--- a/sdk/src/transactions/crypto.rs
+++ b/sdk/src/transactions/crypto.rs
@@ -9,68 +9,88 @@ use crate::{
 /// gas budget before calling this function for rate limiting purposes. Also
 /// rate limited per address.
 pub fn claim_pre_key_for_self(
-    tx: &mut sui::ProgrammableTransactionBuilder,
+    tx: &mut sui::tx::TransactionBuilder,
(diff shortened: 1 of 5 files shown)


sdk/src/transactions/crypto.rs
```diff
@@ -9,68 +9,88 @@ use crate::{
 /// gas budget before calling this function for rate limiting purposes. Also
 /// rate limited per address.
 pub fn claim_pre_key_for_self(
-    tx: &mut sui::ProgrammableTransactionBuilder,
+    tx: &mut sui::tx::TransactionBuilder,
     objects: &NexusObjects,
-) -> anyhow::Result<sui::Argument> {
+) -> anyhow::Result<sui::types::Argument> {
```
nonast, line 14: do we still need to return a result for these functions?

aelesbao: good point; I don't see any `?` in the `fn` anymore.

kouks: They were added back

aelesbao (approved):

PR #333, opened 2025-12-11
grpc: crates, idents and transaction templates

kouks:
## Notable changes

- install required crates
- adjust sui identifiers to use grpc types
- adjust transaction templates to use grpc types

## References

- references https://github.com/Talus-Network/nexus/issues/594

## Checklist

- [x] keep a changelog
- [x] write tests

### DIFF
--- a/sdk/src/idents/mod.rs
+++ b/sdk/src/idents/mod.rs
@@ -8,11 +8,11 @@
 //! use nexus_sdk::sui;
 //! use nexus_sdk::idents::workflow;
 //!
-//! let mut tx = sui::ProgrammableTransactionBuilder::new();
-//! let workflow_pkg_id = sui::ObjectID::random();
+//! let mut tx = sui::tx::TransactionBuilder::new();
+//! let workflow_pkg_id = sui::types::Address::random();
 //! let vertex = workflow::Dag::vertex_from_str(&mut tx, workflow_pkg_id, "my_vertex");
 //!
-//! assert!(matches!(vertex, Ok(sui::Argument::Result(_))));
+//! assert!(matches!(vertex, Ok(sui::types::Argument::Result(_))));
 //! ```
 
 pub mod move_std;
@@ -21,17 +21,30 @@ pub mod sui_framework;
 pub mod tap;
 pub mod workflow;
 
-use crate::sui;
+use {
+    crate::sui,
+    base64::{prelude::BASE64_STANDARD as BASE64, Engine},
+    serde::Serialize,
+};
 
 /// This struct is used to define Nexus Move resources as `const`s.
 pub struct ModuleAndNameIdent {
-    pub module: &'static sui::MoveIdentStr,
-    pub name: &'static sui::MoveIdentStr,
+    pub module: sui::types::Identifier,
+    pub name: sui::types::Identifier,
 }
 
 impl ModuleAndNameIdent {
     /// Returns the fully-qualified string for this identifier under the given package ID.
-    pub fn qualified_name(&self, package: sui::ObjectID) -> String {
+    pub fn qualified_name(&self, package: sui::types::Address) -> String {
         format!("{package}::{}::{}", self.module, self.name)
     }
+}
+
+/// Helper to create a pure [`sui::tx::Input`].
+pub fn pure_arg<T: Serialize>(value: &T) -> anyhow::Result<sui::tx::Input> {
+    Ok(sui::tx::Input {
+        value: Some(sui::tx::Value::String(BASE64.encode(bcs::to_bytes(value)?))),
+        kind: Some(sui::tx::InputKind::Pure),
+        ..Default::default()
+    })
 }
--- a/sdk/src/idents/move_std.rs
+++ b/sdk/src/idents/move_std.rs
@@ -1,30 +1,37 @@
-use crate::{idents::ModuleAndNameIdent, sui};
+use crate::{
+    idents::{move_std, pure_arg, ModuleAndNameIdent},
+    sui,
+};
+
+pub const PACKAGE_ID: sui::types::Address = sui::types::Address::from_static("0x1");
 
 // == `std::ascii` ==
 
 pub struct Ascii;
 
-const ASCII_MODULE: &sui::MoveIdentStr = sui::move_ident_str!("ascii");
+const ASCII_MODULE: sui::types::Identifier = sui::types::Identifier::from_static("ascii");
 
 impl Ascii {
     /// `std::ascii::string`
     pub const STRING: ModuleAndNameIdent = ModuleAndNameIdent {
         module: ASCII_MODULE,
-        name: sui::move_ident_str!("string"),
+        name: sui::types::Identifier::from_static("string"),
     };
 
     /// Convert a string to a Move ASCII string.
     pub fn ascii_string_from_str<T: AsRef<str>>(
-        tx: &mut sui::ProgrammableTransactionBuilder,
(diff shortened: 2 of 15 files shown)


sdk/src/idents/mod.rs
```diff
@@ -21,17 +21,30 @@ pub mod sui_framework;
...
 
-use crate::sui;
+use {
+    crate::sui,
+    base64::{prelude::BASE64_STANDARD as BASE64, Engine},
+    serde::Serialize,
+};
 
 /// This struct is used to define Nexus Move resources as `const`s.
 pub struct ModuleAndNameIdent {
```
davidrotari19, line 31: Does not `StructTag` already does something very similar?

sdk/src/idents/move_std.rs
```diff
@@ -34,46 +41,46 @@ impl Ascii {
...
     /// Convenience helper to build the `std::string::String` type tag.
-    pub fn type_tag() -> sui::MoveTypeTag {
-        sui::MoveTypeTag::Struct(Box::new(sui::MoveStructTag {
-            address: *sui::MOVE_STDLIB_PACKAGE_ID,
-            module: Self::STRING.module.into(),
-            name: Self::STRING.name.into(),
-            type_params: vec![],
-        }))
+    pub fn type_tag() -> sui::types::TypeTag {
+        sui::types::TypeTag::Struct(Box::new(sui::types::StructTag::new(
```
davidrotari19, line 79: Maybe implement `StructTag` and `TypeTag` makes `ModuleAndNameIdent` redundant?

sdk/src/idents/primitives.rs
```diff
@@ -1,90 +1,90 @@
...
     /// `nexus_primitives::data::walrus_many_encrypted`
     pub const WALRUS_MANY_ENCRYPTED: ModuleAndNameIdent = ModuleAndNameIdent {
         module: DATA_MODULE,
-        name: sui::move_ident_str!("walrus_many_encrypted"),
+        name: sui::types::Identifier::from_static("walrus_many_encrypted"),
     };
     /// Create NexusData from a vector of bytes that are stored on Walrus.
     ///
     /// `nexus_primitives::data::walrus_one`
     pub const WALRUS_ONE: ModuleAndNameIdent = ModuleAndNameIdent {
```
davidrotari19, line 68: Could be wrong, but they all look way too similar to `StructTag`

sdk/src/idents/primitives.rs
```diff
@@ -193,41 +203,41 @@ impl OwnerCap {
...
+        name: sui::types::Identifier::from_static("witness_symbol"),
     };
 }
 
-/// Helper to turn a `ModuleAndNameIdent` into a `sui::MoveTypeTag`. Useful for
+/// Helper to turn a `ModuleAndNameIdent` into a `sui::types::TypeTag`. Useful for
 /// creating generic types.
 pub fn into_type_tag(
-    workflow_pkg_id: sui::ObjectID,
+    workflow_pkg_id: sui::types::Address,
```
davidrotari19, line 232: Again I think if you just implement `StructTag`  then conversion to `TypeTag` is trivial

sdk/src/idents/mod.rs
kouks: It does indeed, changing this at the moment would however result in many more replacements in both code bases so I created a ticket here: https://github.com/Talus-Network/nexus-sdk/issues/345

Another thing is that if we swap to your `move_binding` crate, this will be redundant entirely.

sdk/src/idents/primitives.rs
kouks: Yeah all good points - I overlooked this simplification in my refactoring. Like I mentioned on the other comment (and this is likely relevant to all of them) - I will fix separately via the ticket.

sdk/src/idents/sui_framework.rs
```diff
@@ -65,114 +64,114 @@ impl Object {
...
 // == `sui::coin` ==
 
 pub struct Coin;
 
-const COIN_MODULE: &sui::MoveIdentStr = sui::move_ident_str!("coin");
+const COIN_MODULE: sui::types::Identifier = sui::types::Identifier::from_static("coin");
 
 impl Coin {
     /// `sui::coin::Coin`
     pub const COIN: ModuleAndNameIdent = ModuleAndNameIdent {
```
davidrotari19, line 133: I think you have `ModuleAndNameIdent ` because we have both `Structures` and `calls` in the same implementation.  But doesnt sui-rust-sdk exposes `Function` that seems like you make less conversation between types?

sdk/src/idents/workflow.rs
```diff
@@ -18,382 +19,382 @@ impl DefaultTap {
     /// `nexus_workflow::default_tap::begin_dag_execution`
     pub const BEGIN_DAG_EXECUTION: ModuleAndNameIdent = ModuleAndNameIdent {
         module: DEFAULT_TAP_MODULE,
-        name: sui::move_ident_str!("begin_dag_execution"),
+        name: sui::types::Identifier::from_static("begin_dag_execution"),
     };
     /// The witness type needed to register DAG execution.
     ///
     /// `nexus_workflow::default_tap::BeginDagExecutionWitness`
     pub const BEGIN_DAG_EXECUTION_WITNESS: ModuleAndNameIdent = ModuleAndNameIdent {
```
davidrotari19, line 27: I think all of these are `Functions` but maybe what you write makes sense

sdk/src/idents/workflow.rs
```diff
@@ -402,257 +403,275 @@ impl Dag {
...
         let str = super::move_std::Ascii::ascii_string_from_str(tx, str)?;
 
-        Ok(tx.programmable_move_call(
-            workflow_pkg_id,
-            Self::ENTRY_GROUP_FROM_STRING.module.into(),
-            Self::ENTRY_GROUP_FROM_STRING.name.into(),
-            vec![],
+        Ok(tx.move_call(
+            sui::tx::Function::new(
+                workflow_pkg_id,
```
davidrotari19, line 512: I guess ideally the intent should already return the `Function`? seems like we do double work

sdk/src/idents/move_std.rs
davidrotari19: Closed due to your previous comment

sdk/src/idents/primitives.rs
davidrotari19: Closed due to your previous comments

davidrotari19 (approved):
LGTM

sdk/src/idents/workflow.rs
kouks: Yeah, again, good points - I will add `Function` to the ticket.

sdk/src/transactions/scheduler.rs
```diff
@@ -835,115 +914,131 @@ mod tests {
...
         }
 
-        fn commands_len(&self) -> usize {
-            self.tx.commands.len()
+        fn commands(&self) -> &Vec<sui::types::Command> {
+            let sui::types::TransactionKind::ProgrammableTransaction(
+                sui::types::ProgrammableTransaction { commands, .. },
+            ) = &self.tx.kind
+            else {
+                panic!("expected PTB transaction kind, got {:?}", self.tx.kind);
```
nonast, line 933: can this happen, and should it crash here, is it ok to panic? If not, should we use unreachable instead?

kouks: this is test code so it's fine to panic imo, it also won't ever happen

PR #338, opened 2025-12-11
[Preview]: All GRPC-related changes

kouks:


rust-toolchain.toml
```diff
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "1.86"
+channel = "nightly"
```
loverdos, line 2: why?

kouks: `sui-move-build` needs this when bumped to `mainnet-v1.61.2` - it's only needed in tests however, I just wasn't sure how to separate the 2 verisons

PR #344, opened 2026-01-06
[DRAFT] feat: Use `sui move summary` for onchain tool registration

nonast:
## Notable changes

- We now use the summary command to generate a summary of an onchain tool package, which we use to generate input and output schemas from. This is used during onchain tool registration.
- We removed the ability to customize input/output schema arg names. We now use the actual arg names from the summary.
- We also don't use integer names anymore for onchain tool input args.

## References

- fixes https://github.com/Talus-Network/nexus/issues/603
- fixes https://github.com/Talus-Network/nexus/issues/606

## Checklist

- [x] keep a changelog
- [x] write tests


cli/src/tool/tool_register_onchain.rs
```diff
@@ -137,57 +134,51 @@ fn validate_gas_and_collateral_coins(
...
+    let summary_handle = loading!("Running 'sui move summary' on package...");
+    let module_summary = match onchain_schema_gen::run_summary_command(package_path, module_name) {
+        Ok(summary) => {
+            summary_handle.success();
+            summary
         }
         Err(e) => {
-            input_handle.error();
+            summary_handle.error();
             return Err(NexusCliError::Any(e));
```
kouks, line 182: if we go down this route, we should probably return a very granular error here - if `sui` binary is not installed, we should explain why we need it and provide steps to install it. The documentation should also be updated for this command highlighting the dependency on sui cli.

aelesbao: The `summary` command is [publicly accessible](https://github.com/MystenLabs/sui/blob/main/external-crates/move/crates/move-cli/src/base/summary.rs#L60#L5) in the `move-cli`. We could call it directly instead of relying on the `sui` CLI being installed (even thought it's extremely unusual for someone to not have it installed when writing Move contracts).

sdk/src/onchain_schema_gen/summary.rs
```diff
@@ -0,0 +1,683 @@
+//! Summary-based schema generation for Move onchain tools.
```
aelesbao, line 1: Can we use the same structs [available in `move_model_2::summary` here](https://github.com/MystenLabs/sui/blob/main/external-crates/move/crates/move-model-2/src/summary.rs) instead of re-defining the whole object tree?

cli/src/tool/tool_register_onchain.rs
nonast: Changed it to call the command directly: [6fa772a](https://github.com/Talus-Network/nexus-sdk/pull/344/commits/6fa772a374805c40f7210bfa401bf02dd40830a6) 
[00dfcd7](https://github.com/Talus-Network/nexus-sdk/pull/344/commits/00dfcd72c9dd6252284b6f60b8f0dd1760ab1fe6)

sdk/src/onchain_schema_gen/summary.rs
nonast: Yes, nice one: [8f2c9e3](https://github.com/Talus-Network/nexus-sdk/pull/344/commits/8f2c9e3c3195e718015d975934d087a7807feb57)

nonast:
After the discussion yesterday it became apparent that directly calling the summary command in sui's "external" crate is not wanted due our shift to sui's rust-sdk. There are a few realistic options I can think of now: 

1. We write our own lightweight "summary generator". It will just be a simple source code parser, since you only need the Output enum and execute function signature.
2. Use MovePackageService grpc service to introspect this info from the published package (similar to original approach, but shifted to grpc).

There are other options but left them out of the picture since they don't seem that user friendly.

kouks:
Is this PR irrelevant now @nonast ?

PR #361, opened 2026-01-08
[0.5.0] refactor: Unify tool info and locations for both tool variants

nonast:
## Notable changes

- added `ToolLocation` to unify `ToolInfo`

## References

- fixes https://github.com/Talus-Network/nexus/issues/588

## Checklist

- [x] keep a changelog
- [x] write tests

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -20,13 +20,13 @@ Set of commands for managing Tools.
 
 ---
 
-**`nexus tool new <name> --template <template>`**
+**`nexus tool new --name <name> --template <template>`**
 
 Create a new Tool scaffolding in a folder called `<name>`. Which files are generated is determined by the `--template` flag. I propose having `templates/tools/<template>.template` files that contain the Tool skeleton files. For example for `rust` it'd be a `Cargo.toml` with the `nexus-toolkit` dependency, and a `src/main.rs` file that shows a basic use case of the crate.
 
 ---
 
-**`nexus tool validate --off-chain <url>`**
+**`nexus tool validate off-chain --url <url>`**
 
 Validate an off-chain Nexus Tool on the provided URL. This command checks whether the URL hosts a valid Nexus Tool interface:
 
@@ -42,7 +42,7 @@ This command should also check that the URL is accessible by the Leader node. It
 
 ---
 
-**`nexus tool validate --on-chain <ident>`**
+**`nexus tool validate on-chain --ident <ident>`**
 
 {% hint style="warning" %}
 The specific design for onchain tools is still in progress and as a result the implementation is not yet present. When running the command, it will panic.
@@ -50,7 +50,7 @@ The specific design for onchain tools is still in progress and as a result the i
 
 ---
 
-**`nexus tool register --off-chain <url> --invocation-cost [mist] --collateral-coin [object_id] [--batch] [--no-save]`**
+**`nexus tool register offchain --url <url> --invocation-cost [mist] --collateral-coin [object_id] [--batch] [--no-save]`**
 
 Command that makes a request to `GET <url>/meta` to fetch the Tool definition and then submits a TX to our Tool Registry. It also locks the collateral and sets the single invocation cost of the Tool which defaults to 0 MIST.
 
@@ -70,7 +70,7 @@ Tool registration is currently restricted during the beta phase. To register you
 
 ---
 
-**`nexus tool register --on-chain <ident>`**
+**`nexus tool register on-chain --package <ADDRESS> --module <MODULE> --tool-fqn <FQN> --description <DESCRIPTION> --witness-id <OBJECT_ID>`**
 
 {% hint style="warning" %}
 The specific design for onchain tools is still in progress and as a result the implementation is not yet present. When running the command, it will panic.
--- a/docs/guides/setup.md
+++ b/docs/guides/setup.md
@@ -51,7 +51,7 @@ nexus --version
 ## Download the Nexus objects
 
 ```bash
-wget https://storage.googleapis.com/production-talus-sui-objects/v0.4.0/objects.devnet.toml
+wget -O ~/.nexus/objects.devnet.toml https://storage.googleapis.com/production-talus-sui-objects/v0.4.0/objects.devnet.toml
 ```
 
 ## Configure the Talus devnet
@@ -59,9 +59,10 @@ wget https://storage.googleapis.com/production-talus-sui-objects/v0.4.0/objects.
 Configure your Nexus CLI to connect to the Talus `devnet` by running:
 
 ```bash
-nexus conf set --sui.net devnet \
-  --sui.rpc-url https://rpc.ssfn.devnet.production.taluslabs.dev \
-  --nexus.objects objects.devnet.toml
+nexus conf set \
+  --sui.rpc-url https://grpc.ssfn.devnet.production.taluslabs.dev \
+  --sui.gql-url https://graphql.devnet.production.taluslabs.dev/graphql \
+  --nexus.objects ~/.nexus/objects.devnet.toml
 ```
 
 ### Configure the Sui client
@@ -94,6 +95,14 @@ sui client switch --address tally
 {% hint style="danger" %}
 This command will output your wallet details, including your address and recovery phrase. Ensure you store this information securely.
 {% endhint %}
+
+Import the newly created wallet to `nexus`:
+
+```bash
+PK=$(sui keytool export --key-identity tally --json | jq -er '.exportedPrivateKey')
+BASE64_PK=$(sui keytool convert "$PK" --json | jq -er '.base64WithFlag')
+nexus conf set --sui.pk "$BASE64_PK"
+```
 
 To request funds from the faucet, run the following command twice to get 2 gas coins:
 


CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - support for `scheduler` transactions and events
 - onchain schema generation
+- `ToolLocation` to combine offchain url and onchain movemoduleid
```
davidrotari19: MoveModuleId or move_module_id? seems better to read, but if you think this is better resolve

cli/src/tool/tool_list.rs
```diff
@@ -7,6 +7,7 @@ use {
             deserialize_bytes_to_url,
             deserialize_string_to_datetime,
         },
+        ToolLocation,
```
davidrotari19: ToolIdentifier? why location isnt that a bit confusing? 
- Location means its different from idenitity but we treat them the same
- This seems more like a reference or a pointer so not a location

sdk/src/tool_ref.rs
```diff
@@ -0,0 +1,296 @@
...
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum ToolLocation {
+    /// HTTP(S) endpoint for offchain tools.
+    Http(reqwest::Url),
+    /// Sui module for onchain tools (package::module format).
+    Sui {
+        /// The package address containing the tool module.
+        package: sui::types::Address,
+        /// The module name within the package.
+        module: sui::types::Identifier,
```
loverdos, line 38: I believe I've seen a `ModuleId` struct somewhere in the Sui types, which is exactly this combination of `Address` and `Identifier`. If we can find it we should use it.

cli/src/tool/tool_list.rs
```diff
@@ -51,47 +52,62 @@ pub(crate) async fn list_tools() -> AnyResult<(), NexusCliError> {
...
+            ),
+            ToolVariant::OnChain(t) => (
+                // Parse package address and create Sui location.
+                ToolLocation::new_sui(
+                    t.package_address
+                        .parse()
+                        .unwrap_or_else(|_| sui::types::Address::ZERO),
+                    sui::types::Identifier::new(&t.module_name)
+                        .unwrap_or_else(|_| sui::types::Identifier::from_static("unknown")),
+                ),
```
loverdos: What's the purpose of filling in `ZERO` and `"Unknown"` instead of just failing?

Also, worth propagating this parsing logic to `ToolLocation` itself.

cli/src/tool/tool_list.rs
```diff
@@ -51,47 +52,62 @@ pub(crate) async fn list_tools() -> AnyResult<(), NexusCliError> {
...
-                    "OnChain Tool '{fqn}' at '{package}::{module}' registered '{registered_at}' - {description}",
-                    fqn = fqn.to_string().truecolor(100, 100, 100),
-                    package = onchain_tool.package_address.truecolor(100, 100, 100),
-                    module = onchain_tool.module_name.truecolor(100, 100, 100),
-                    registered_at = onchain_tool.registered_at_ms.to_string().truecolor(100, 100, 100),
-                    description = onchain_tool.description.truecolor(100, 100, 100),
-                );
-            }
+        let (location, description, registered_at_ms, witness_id, input_schema) = match &tool {
+            ToolVariant::OffChain(t) => (
```
kouks, line 56: This is probably for a ticket and a future PR _but_ - I find this double separation convoluted. We use both the `ToolVariant` and `ToolLocation` to separate onchain/offchain - why?

If `ToolVariant` holds the location and has `is_onchain` function then the wrapper type `ToolVariant` shouldn't be an enum. Or vice versa. I know this depends on the tool registry types _but_ - those should be adjusted too to follow a similar structure. So ideally we'd have:

```rs
/// Holds the generic tool info.
struct Tool {
  /// Location is what differentiates on/offchain.
  location: ToolLocation,
  description,
  registered_at_ms,
  input_schema,
  output_schema,
}
```

cli/src/tool/tool_list.rs
```diff
@@ -51,47 +52,62 @@ pub(crate) async fn list_tools() -> AnyResult<(), NexusCliError> {
...
+                    t.package_address
+                        .parse()
+                        .unwrap_or_else(|_| sui::types::Address::ZERO),
+                    sui::types::Identifier::new(&t.module_name)
+                        .unwrap_or_else(|_| sui::types::Identifier::from_static("unknown")),
+                ),
+                t.description.clone(),
+                t.registered_at_ms,
+                Some(t.witness_id.clone()),
+                Some(t.input_schema.clone()),
```
kouks: Why is `input_schema` optional and only present for onchain tools? Don't both types of tools have it? Also where is `output_schema`?

cli/src/tool/tool_list.rs
```diff
@@ -51,47 +52,62 @@ pub(crate) async fn list_tools() -> AnyResult<(), NexusCliError> {
...
+                ToolLocation::new_sui(
+                    t.package_address
+                        .parse()
+                        .unwrap_or_else(|_| sui::types::Address::ZERO),
+                    sui::types::Identifier::new(&t.module_name)
+                        .unwrap_or_else(|_| sui::types::Identifier::from_static("unknown")),
+                ),
+                t.description.clone(),
+                t.registered_at_ms,
+                Some(t.witness_id.clone()),
```
kouks: Since `witness_id` is specific to on-chain tools - should this be on the `ToolLocation` struct, too?

cli/src/tool/tool_list.rs
```diff
@@ -51,47 +52,62 @@ pub(crate) async fn list_tools() -> AnyResult<(), NexusCliError> {
...
+            "registered_at_ms": registered_at_ms,
+            "description": description,
+        });
+
+        if let Some(wid) = &witness_id {
+            tool_json["witness_id"] = json!(wid);
         }
+        if let Some(schema) = &input_schema {
+            tool_json["input_schema"] = json!(schema);
+        }
```
kouks: Ideally the `ToolVariant` type would be `Serialize` and this could all be avoided.

CHANGELOG.md
kouks: or just "move module ID" it is supposed to just be a readable text

nonast: [4ceee86](https://github.com/Talus-Network/nexus-sdk/pull/361/commits/4ceee86f77ceadd5d488b371838282963d6d0b11)

also moved to 0.5.0

cli/src/tool/tool_list.rs
nonast: Identity could cause confusion with the move `Identitifier` which is just the module name. Tool location should encapsulate both the move package + indentifier, and offchain url. 

ToolLocation is slightly derived from the new rust sdk struct `MoveLocation`, which isn't exactly what `MoveModuleId` used to be in the previous sdk, but still contains package + module. (It's more detailed and used for aborts now).

sdk/src/tool_ref.rs
nonast: Correct, we used this struct previously, but since the move to the new sdk, this struct disappeared. 

It was a nice little struct, you shall be missed.

cli/src/tool/tool_list.rs
nonast: Agreed, better to fail than show fake info. Also moved parsing. [73fe985](https://github.com/Talus-Network/nexus-sdk/pull/361/commits/73fe985dc11e75ad7044b262ed72e675327ed2d5)

cli/src/tool/tool_list.rs
nonast: https://github.com/Talus-Network/nexus/issues/650

cli/src/tool/tool_list.rs
nonast: I think I put in as test once for onchain tools, but never changed it. I've added them both for both variants now and removed the option. [7d48cf6](https://github.com/Talus-Network/nexus-sdk/pull/361/commits/7d48cf6dc4078cad2e30116e151092326b59c2ef)

Tested and shows correct fields.

cli/src/tool/tool_list.rs
nonast: Yes, I didn't add this in previously since we had MoveModuleID which was 0x1::module, so parsing there to toollocation was easy. But it didn't have witness id, so it needed to be added as "extension", which didn't look that nice. 

But since we don't do that anymore I think it makes sense to put it in there as well. Added it in the same string to keep the string serialization intact.

One extra change I made now was in deserialization of the onchain tool, we didn't retrieve the witness id from onchain because the field was missing in the OnChainTool struct used for deserializing. Didn't notice that because it wasn't really needed in the backend. Witness id isn't used at all, only on chain.

[eeb80b7](https://github.com/Talus-Network/nexus-sdk/pull/361/commits/eeb80b727639468c38f17a8e0efb509f699fb0d6)
[40cd4ba](https://github.com/Talus-Network/nexus/pull/621/commits/40cd4ba684a85c531f8659e6e1bf3f01f33d1068)

cli/src/tool/tool_list.rs
nonast: We're adding fqn and tooltype to it which isn't in toolvariant. Also, I added a few other fields to make toollocation a bit clearer. I don't think simply adding serialize to toolvariant would make it a lot clearer. But I could be wrong.

cli/src/tool/tool_list.rs
davidrotari19: The problem is that url is not a location not package::module. Url is a reference and package::module is an onchain identifier(as pacakges are objects) so location means a physical location which none is. A better name if ToolIdentifier is not good(in my opinion it is fine). ToolReference

nonast: How about ToolRef

cli/src/tool/tool_list.rs
loverdos: yes I agree with @kouks

cli/src/tool/tool_list.rs
kouks: We can review this as part of the tool* ticket

PR #362, opened 2026-01-11
feat(cli): enable global `--verbose` log

aelesbao:
## Notable changes

- Initialize the `env_logger` crate to debug operations in the `nexus` CLI

## Checklist

- [x] keep a changelog
- [ ] write tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - `nexus scheduler task create` / `inspect` / `metadata` / `pause` / `resume` / `cancel`
   - `nexus scheduler occurrence add`
   - `nexus scheduler periodic set` / `disable`
-- enable more verbose output using the `RUST_LOG` env var
+- add `--verbose` flag for debug log output
 
 ### `nexus-sdk`
 


rust-toolchain.toml
```diff
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "nightly"
+channel = "nightly-2026-01-01"
```
loverdos: This alone won't work because we go to [great lengths](https://github.com/Talus-Network/nexus-sdk/blob/a73ea788e87ce30b30fe9a17c184f4d88eda52dd/.pre-commit/.just#L230-L273) to ensure we are being consistent about the _exact_ nightly version.

But the more important question is why do we need `nightly`? Our goal should be to eliminate nightly altogether not to propagate it further. 

See also https://github.com/Talus-Network/nexus-sdk/pull/365

cli/src/main.rs
```diff
@@ -52,6 +53,8 @@ enum Command {
 
 #[tokio::main]
 async fn main() {
+    env_logger::init();
```
kouks: would we want to hide this behind some `--verbose` global arg like `--json`?

aelesbao: [d3f22b0](https://github.com/Talus-Network/nexus-sdk/pull/362/commits/d3f22b054c6337d5a64da0e251de1666d68ed8e0)

rust-toolchain.toml
aelesbao: Reverted [a7b936a](https://github.com/Talus-Network/nexus-sdk/pull/362/commits/a7b936a5035cd3daf332a74d81ffbe899e2ced96)

PR #374, opened 2026-01-19
[0.6.0] feat: use `SharedObjectRef` in announce event

kouks:
## Notable changes

- see `CHANGELOG.md`

## References

- part of https://github.com/Talus-Network/nexus/issues/651

## Checklist

- [x] keep a changelog
- [x] write tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Added
 
-- `AnnounceInterfacePackageEvent.objects_by_ref_mut` field to indicate if shared objects are passed by mutable reference
+- `SharedObjectRef` type to represent shared object references with mutability information
+- `AnnounceInterfacePackageEvent` now has `shared_objects` field of type `Vec<SharedObjectRef>` that carries the reference type information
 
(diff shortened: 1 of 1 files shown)


sdk/src/types/shared_object_ref.rs
```diff
@@ -0,0 +1,73 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct SharedObjectRef {
+    pub id: sui::types::Address,
+    pub ref_mut: bool,
+}
+
+impl SharedObjectRef {
+    /// Create a new immutable shared object reference.
+    pub fn new_imm(id: sui::types::Address) -> Self {
```
aelesbao, line 17: We could use the same semantics as Rust for `let/let mut` and name this function `new` (and keep the other as `new_mut`)

aelesbao (approved):

sdk/src/types/shared_object_ref.rs
kouks: I have mixed feelings about this - it would match up with rust but on the other hand the interface could be slightly confusing. I'd expect `new_mut` to do something more than just `new` - as in I'd expect `new` to take the mutability as an argument, too.

PR #396, opened 2026-02-10
[0.6.0] feat: leader distribution support

kouks:
## Notable changes

- see `CHANGELOG.md`

## References

- part of https://github.com/Talus-Network/nexus/pull/679/changes

## Checklist

- [x] keep a changelog
- [x] write tests

### DIFF
--- a/sdk/src/types/serde_parsers.rs
+++ b/sdk/src/types/serde_parsers.rs
@@ -304,7 +304,7 @@ where
 
 #[cfg(test)]
 mod tests {
-    use {super::*, serde::Deserialize};
+    use {super::*, chrono::TimeZone, serde::Deserialize};
 
     #[derive(Deserialize, Serialize)]
     struct TestUrlStruct {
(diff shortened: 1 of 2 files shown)


sdk/src/events/mod.rs
```diff
@@ -10,6 +10,31 @@ mod parsing;
...
+/// to execute the event.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct DistributedEventMetadata {
+    /// The execution window duration.
+    #[serde(
+        rename = "deadline_ms",
+        deserialize_with = "deserialize_sui_u64_to_duration",
+        serialize_with = "serialize_duration_to_sui_u64"
+    )]
+    pub deadline: chrono::Duration,
```
davidrotari19, line 24: I think time here is `sui` time, which isnt `Duration`? I think it is `u64`

sdk/src/events/mod.rs
```diff
@@ -10,6 +10,31 @@ mod parsing;
 
 pub use {fetching::*, graphql::*, parsing::*};
 
+/// Distribution metadata for distributed events. This contains metadata about
+/// the event deadline as well as the priority in which leaders should attempt
```
davidrotari19, line 14: Does it have priority?

sdk/src/events/mod.rs
kouks: that's why the custom deser - it converts serialized `u64` as string to `u64` and then to `Duration`

sdk/src/events/mod.rs
kouks: it has the `leaders` vector which is sorted based on the onchain algo

sdk/src/events/parsing.rs
```diff
@@ -236,6 +248,24 @@ mod tests {
         event: T,
     }
 
+    #[derive(Clone, Debug, Serialize, Deserialize)]
+    struct DistributedWrapper<T> {
+        event: T,
+        deadline_ms: String,
```
davidrotari19, line 254: Is this for `JSON`?

kouks: yeah graphql JSON has `u64` as string

sdk/src/types/derive.rs
```diff
@@ -58,6 +59,46 @@ pub fn derive_execution_gas_id(
     derive_object_id(gas_service, &tag, &execution)
 }
 
+/// Derive the object ID for a [`RequestWalkExecutionEvent`] task given the
+/// execution ID and the [`RuntimeVertex`].
+///
+/// Note that this ID derived twice - first type based on the vertex name and
+/// second time based on the iteration.
```
davidrotari19, line 66: Isnt it just on (vertex_name, iteration)? so its not twice?

sdk/src/types/serde_parsers.rs
```diff
@@ -192,6 +192,31 @@ where
...
+/// Deserialize a duration in milliseconds stored as a string.
+pub fn deserialize_sui_u64_to_duration<'de, D>(
+    deserializer: D,
+) -> Result<chrono::Duration, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let millis = deserialize_sui_u64(deserializer)?;
+
+    Ok(chrono::Duration::milliseconds(millis as i64))
```
davidrotari19, line 204: Is this better than just keeping it u64

sdk/src/types/derive.rs
kouks: execution ID + vertex_name = some_id
some_id + iteration = resulting_id

It is derived twice - question is if it can be improved to only derive once.

sdk/src/events/mod.rs
davidrotari19: You mean leader priority

sdk/src/types/derive.rs
davidrotari19: why not execution ID + (vertex_name + iteration)?

sdk/src/types/serde_parsers.rs
kouks: I think that's more of a philosophical question - I don't know. I use it as `chrono::Duration` in leader and you can always get the ms back.

Semantically it feels better because it's not just a magic number but a struct describing its purpose. Especially when you have `deadline_ms` and `requested_at_ms` - it's nice that it's parsed into `Duration` and `DateTime<UTC>` to clarify what is a timestamp and what is not.

sdk/src/types/derive.rs
davidrotari19: You are doing this, but I meant (vertex_name, iteration) this tuple is unique so you can use it directly

davidrotari19: Maybe in another pr

kouks: Mainly because I don't know how a `sui::types::TypeTag` works for a tuple of `(AsciiString, u64)`.

kouks: I was thinking about converting both to a `vector<u8>` (and concating) but that felt worse (at least in terms of readability) than deriving twice.

kouks: > Maybe in another pr

sounds good

PR #462, opened 2026-05-25
update for scheduler

LKY-stephen:
## Notable changes

 - Updates SDK TAP/registry/scheduler idents,
  transaction builders, types, event parsing, Nexus
  object metadata, and CLI TAP/tool commands.

## Reviewer Focus

- SDK/leader alignment for endpoint revision events,
  active skill revision updates, scheduled TAP task
  recovery, and execution payment settlement.

## Checklist

- [x] keep a changelog
- [ ] write tests


sdk/src/types/nexus_objects.rs
```diff
@@ -18,15 +20,20 @@ use {
...
     pub network_id: sui::types::Address,
     pub registry_pkg_id: sui::types::Address,
     pub tool_registry: sui::types::ObjectReference,
     pub verifier_registry: sui::types::ObjectReference,
     pub network_auth: sui::types::ObjectReference,
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub tap_registry: Option<sui::types::ObjectReference>,
+    #[serde(
+        default,
+        alias = "tap_registry",
```
kouks: Do we need this alias here? I'd rather keep it consistent since we don't need backwards compatibility for this yet.

sdk/src/types/nexus_objects.rs
```diff
@@ -77,10 +90,20 @@ impl NexusObjects {
...
+    pub fn scheduler_type_origin_pkg_id(&self) -> sui::types::Address {
+        self.scheduler_original_pkg_id
+            .unwrap_or(self.scheduler_pkg_id)
+    }
+
     /// Returns the shared standard TAP registry object reference when the
     /// deployment metadata includes it.
-    pub fn tap_registry(&self) -> Option<&sui::types::ObjectReference> {
-        self.tap_registry.as_ref()
+    pub fn agent_registry(&self) -> Option<&sui::types::ObjectReference> {
```
kouks: Do we need this function? None of the other fields have this getter.

cli/src/tap/tap_output.rs
```diff
@@ -24,18 +24,12 @@ pub(crate) fn publish_skill_result_json(result: &PublishSkillResult) -> serde_js
         "dag_id": result.dag.dag_object_id,
         "dag_digest": result.dag.tx_digest,
         "dag_checkpoint": result.dag.tx_checkpoint,
-        "endpoint_object_id": result.endpoint.endpoint_object.object_id(),
-        "endpoint_object_version": result.endpoint.endpoint_object.version(),
-        "endpoint_object_digest": result.endpoint.endpoint_object.digest(),
-        "endpoint_digest": result.endpoint.tx_digest,
-        "endpoint_checkpoint": result.endpoint.tx_checkpoint,
```
kouks, line 31: Are we dropping the endpoint idea altogether?

sdk/src/events/mod.rs
```diff
@@ -318,12 +302,29 @@ pub struct RequestWalkStandardTapContext {
...
+        }
+    }
+}
+
 impl RequestWalkExecutionEvent {
+    pub fn endpoint_key(&self) -> Option<TapEndpointKey> {
+        Some(TapEndpointKey {
+            agent_id: self.tap_agent_id?,
+            skill_id: self.tap_skill_id?,
+            interface_revision: self.tap_interface_revision?,
```
kouks, line 320: With the latest changes:

> Updates default DAG execution to resolve through agent_registry and default_tap_target

Shouldn't these fields be always present?

sdk/src/events/mod.rs
```diff
@@ -490,6 +477,30 @@ pub struct VertexAuthorizationGrantRequiredEvent {
...
 
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct ScheduledAuthorizationGrantCreatedEvent {
+    pub scheduled_grant_id: sui::types::Address,
+    pub scheduled_task_id: sui::types::Address,
+    pub agent_id: AgentId,
+    pub skill_id: SkillId,
+    pub dag_id: sui::types::Address,
+    #[serde(deserialize_with = "deserialize_bytes_to_string")]
+    pub vertex: String,
```
kouks, line 488: Should the `vertex` fields on these 2 events be `RuntimeVertex` or do we not care about iterations when looping the same vertex?

sdk/src/events/parsing.rs
```diff
@@ -491,6 +491,7 @@ fn is_nexus_package(address: sui::types::Address, objects: &NexusObjects) -> boo
     address == objects.primitives_pkg_id
         || address == objects.interface_pkg_id
         || address == objects.registry_pkg_id()
+        || objects.is_scheduler_package(address)
         || objects.is_workflow_package(address)
```
kouks, line 495: Not a fan of this incosistency. We check in 3 different ways for (almost) the same thing. Twice via a field comparison, once via a return value comparison and twice using a `-> bool` function.

Maybe having a function like `is_nexus_package` would be useful. Or just make it consistent.

cli/src/tap/tap_output.rs
LKY-stephen: The concept still exists, just remove these fields for simplicity, and store it completely inside the agent_registry

sdk/src/events/mod.rs
LKY-stephen: What do you mean? The default DAG execution just uses a default agent with a default skill. The agent is stored inside the agent_registry that controls its access.

sdk/src/events/mod.rs
LKY-stephen: The vertex needs to consume the cap, so if a vertex needs a grant, its signature starts with a cap. Therefore, iteration or not will not affect it, a vertex that needs grant always needs the grant. But regarding whether it is a string or RuntimeVertex, the main reason is for deserialization.

kouks: So if we expect to loop a vertex N times, we need N grants? And it doesn't matter in which order they're used?

sdk/src/events/mod.rs
kouks: Mostly wondering if it'd be better to always populate the `tap_agent_id`, `tap_skill_id` and `tap_interface_version` with some references to the default agent and default skill. But we can keep this for later.

LKY-stephen: Uh, I see your point. This is a general event, so we can use a reserved event for the default one later. I would leave this to the tasks that store more information in the skill records (Endpoint) to simplify the invocation

sdk/src/events/mod.rs
LKY-stephen: Yes, because the grant cap only authenticate the vertex is executed. Under the current design, it should use the input data to distinguish different invocations.

sdk/src/events/parsing.rs
LKY-stephen: This is a bit tricky because of the old original_pkg_id issue. We haven't decided how to split them. I removed the getter, but kept the is_xxx_package for future updates.

sdk/src/types/nexus_objects.rs
LKY-stephen: removed

sdk/src/types/nexus_objects.rs
LKY-stephen: removed

kouks (approved):

cli/src/tap/mod.rs
```diff
@@ -724,21 +669,18 @@ mod tests {
         let register_error = handle(TapCommand::RegisterSkill {
             artifact: artifact_path.clone(),
             agent_id: sui::types::Address::from_static("0xa"),
-            endpoint_object_id: None,
             gas: gas_args(),
         })
         .await
-        .expect_err("register dispatch requires endpoint metadata");
+        .expect_err("register dispatch reaches missing RPC");
```
davidrotari19, line 675: The error here seems wrong?

sdk/src/events/mod.rs
```diff
@@ -490,6 +477,30 @@ pub struct VertexAuthorizationGrantRequiredEvent {
     pub skill_id: Option<SkillId>,
 }
 
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct ScheduledAuthorizationGrantCreatedEvent {
+    pub scheduled_grant_id: sui::types::Address,
+    pub scheduled_task_id: sui::types::Address,
+    pub agent_id: AgentId,
+    pub skill_id: SkillId,
+    pub dag_id: sui::types::Address,
```
davidrotari19, line 486: Dag seems redundant here as skill already points at one?

davidrotari19 (approved):

sdk/src/events/mod.rs
LKY-stephen: Skill may be updated to use a different dag during revision.

cli/src/tap/mod.rs
LKY-stephen: No, the prior endpoint_id requirement is removed for simplicity, so this error returns to the missing RPC one. (means cli passed if my understanding is correct)

PR #480, opened 2026-06-12
fix: reject tool version 0

### LINKED ISSUE #475: Reject `0` as tool version in FQN
At the moment we state we accept "positive u32 integers" but `0` is accepted.

kouks:
## Notable changes

- see `CHANGELOG.md`

## References

- resolves #475 

## Checklist

- [x] keep a changelog
- [x] write tests


loverdos (approved):

sdk/src/tool_fqn.rs
```diff
@@ -108,6 +108,10 @@ impl FromStr for ToolFqn {
             anyhow::bail!("Tool version too large");
         };
 
+        if version == 0 {
+            anyhow::bail!("Tool version cannot be zero");
```
LKY-stephen, line 112: The move code has been fixed?

kouks: No this is for the Rust code only. We have to come up with something more sophisticated for the Move code as there is currently no enforcement on FQN structure at all.

LKY-stephen: Then, this is just preventing the current leader from stopping? Because the onchain logic still allows this, and can keep firing execution that cannot be handled?

kouks: This is more of a cosmetic fix - the documentation says version is a positive u32 but the logic didn't enforce it. This doesn't really have any implications for the leader or the Move code.

LKY-stephen: They why we add the patch?

LKY-stephen: I mean if the move accept but the leader reject, it only cause inconsistency right?

kouks: Move doesn't verify FQNs at all - it's something we should add. I just want to make the SDK part consistent with documentation.

LKY-stephen: Yes, I know. My point is either you add two together, or you don't add both and update doc. Otherwise, it only introduces a broken behavior.

kouks: https://github.com/Talus-Network/nexus/issues/815 We'll add the Move side as well

LKY-stephen: Uh, sorry, this is just sdk. I thought it was a leader. Not perfect, but not a big deal anyway. You can add an issue for this.

LKY-stephen (approved):

davidrotari19 (approved):

PR #482, opened 2026-06-17
feat: simplify interfaces

### LINKED ISSUE #478: Reorganize Interface and Primitives
1. Relocate definitions that general methodologies to Primitives and Project instantiated object methodologies to Interface
2. Use primitives and interfaces to reduce definition in workflow, registry, and scheduler.

LKY-stephen:
## Notable changes

- Move module alignment: verify `idents::{tap, workflow, registry, scheduler}` and every type tag/call target follow the split Move modules: `agent`, `authorization`, `payment`, `scheduled_request`, `verifier`, registry, scheduler, workflow, and registered-key verifier surfaces.
- Stale interface removal: ensure active SDK/CLI paths no longer expose endpoint revision announcement, config-digest requirements, shared-object requirement plumbing, vertex authorization schema registration, payment-source hashes, registry worksheet confirmation, TAP-specific leader stamp helpers, or `leader_stamp_worksheet_for_dry_run`.
- TAP models: check `TapPublishArtifact`, `TapSkillRequirements`, `TapPaymentPolicy`, `TapSchedulePolicy`, `FixedTool`, current skill records, default-agent/default-executor models, execution context, and payment models reflect the simplified on-chain field set and do not silently invent removed fields.
- Authorization and payment types: verify `sdk/src/types/authorization.rs`, `sdk/src/types/payment.rs`, `sdk/src/types/tap.rs`, and `sdk/src/types/scheduler.rs` preserve grant/cap identity, direct `UserFunded` and `AgentFunded` source identity, scheduled reserve fields, receipts, occurrence state, and error handling.
- PTB builders: verify `transactions::tap`, `transactions::dag`, and `transactions::scheduler` build arguments in the exact Move order for fixed-tool registration, direct execution, scheduled task payment creation, DAG-owned worksheet submission, generic leader worksheet stamping, payment accomplishment/refund, scheduled reserve occurrence completion, and verifier evidence.
- High-level actions: verify `TapActions`, `WorkflowActions`, `SchedulerActions`, crawler helpers, and dynamic-field fetches decode current objects and return explicit `NexusError` variants for missing/malformed state rather than panicking or falling back to stale shapes.
- Event parsing: verify all current TAP, scheduler, payment reserve, workflow request, execution finished, abort, verifier, and grant/cap events decode current BCS/Move option layouts and preserve `scheduled_task_id` where applicable.
- CLI behavior: verify `tap schedule-task`, scheduler task commands, TAP publish/register/update/execute/dry-run/payment/vault flows, command JSON, docs, and templates expose `scheduled_task_id`, current-skill requirements, fixed tools, direct payment modes, and unified direct/scheduled execution without keeping removed attach-style commands or stale aliases.
- Cross-repo coherence:  every SDK active API should target the current Move ABI and any intentionally retained historical decoding should be isolated from active builders/actions.

- see changelog

## References

## Checklist

- [ ] keep a changelog
- [ ] write tests


cli/src/tap/mod.rs
```diff
@@ -341,230 +334,108 @@ pub(crate) enum TapCommand {
...
-            help = "Scheduler task object ID to attach the scheduled TAP task to.",
-            value_name = "OBJECT_ID"
-        )]
-        scheduler_task_id: sui::types::Address,
-        #[arg(long, help = "On-chain generated agent ID.", value_name = "OBJECT_ID")]
+    ScheduledTask(ScheduledTaskCommand),
+}
+
+#[derive(Subcommand)]
+pub(crate) enum ScheduledTaskCommand {
```
kouks, line 415: So now we have 3 potential entry points to an execution?

```
nexus dag execute
nexus scheduler task create
nexus tap schedule-task
```

cli/src/tap/tap_scheduled_task.rs
```diff
@@ -0,0 +1,108 @@
...
+        notify_success,
+        prelude::*,
+        sui::get_nexus_client,
+    },
+    nexus_sdk::nexus::tap::{AgentTaskStateAction, SetAgentTaskStateResult},
+    serde_json::json,
+};
+
+#[derive(Clone, Copy, Debug)]
+pub(crate) enum ScheduledTaskStateRequest {
```
kouks: why duplicate this enum and not reuse `AgentTaskStateAction`?

cli/src/tap/tap_scaffold.rs
kouks: Eventually it'd be nice to extract these to `.jinja` templates as well, akin to `tool.move.jinja` etc.

sdk/src/events/mod.rs
```diff
@@ -216,6 +215,28 @@ where
     })
 }
 
+fn deserialize_move_option_u64<'de, D>(
```
kouks: Would be nice to have this among the other parsers. Also - I thought you'd added one like this already?

sdk/src/events/parsing.rs
kouks, line 1740: Aren't all of these tests tautological? We serialize BCS on the concrete type to ensure it deserializes. But if they serialize, they must deserialize.

That's why I was using the raw bytes taken from on-chain for the `sample_events` below. The main benefit of having the raw bytes is that you cannot swap the order of the fields on the events by accident. Whereas if you serialize->deserialize, you can change the order and the test will still pass.

sdk/src/nexus/crawler.rs
```diff
@@ -706,6 +706,77 @@ impl Crawler {
             .ok_or_else(|| anyhow!("Dynamic object field not found for parent '{parent_id}'"))
     }
 
+    /// Fetch dynamic object fields for a parent and keep only entries whose
+    /// key BCS decodes as `K`. Parents may legitimately contain several child
```
kouks: There is no `K`.

Also - we have an e2e test suite for the crawler `object_crawler_tests.rs`. Would be nice to add an assertion for this to both verify it and present the use case.

sdk/src/onchain_schema_gen/types.rs
```diff
@@ -192,8 +189,20 @@ pub fn is_workflow_vertex_authorization_cap_param(
...
+    if struct_tag.module().as_str() != "authorization"
+        || struct_tag.name().as_str() != "ProvenValue"
+    {
+        return false;
+    }
+    let Some(sui::types::TypeTag::Struct(inner)) = struct_tag.type_params().first() else {
+        return false;
+    };
+    inner.module().as_str() == "authorization"
+        && inner.name().as_str() == "AgentVertexAuthorization"
```
kouks: Both `ProvenValue` and `AgentVertexAuthorization` should probably be added to the `idents` module to avoid str comparison.

sdk/src/transactions/dag.rs
```diff
@@ -3065,6 +2967,11 @@ mod tests {
         let __ph_5 = tx.pure(&5u64);
         let __ph_6 = tx.pure(&6u64);
         let __ph_7 = tx.pure(&7u64);
+        let __ph_7 = tx.pure(&7u64);
+        let __ph_7 = tx.pure(&7u64);
+        let __ph_7 = tx.pure(&7u64);
+        let __ph_7 = tx.pure(&7u64);
+        let __ph_7 = tx.pure(&7u64);
```
kouks: Is this some testing artifact?

sdk/src/transactions/dag.rs
```diff
@@ -3133,6 +3040,8 @@ mod tests {
         let __ph_4 = tx.pure(&4u64);
         let __ph_5 = tx.pure(&5u64);
         let __ph_6 = tx.pure(&6u64);
+        let __ph_7 = tx.pure(&7u64);
+        let __ph_7 = tx.pure(&7u64);
```
kouks: same here

sdk/src/sui.rs
```diff
@@ -9,6 +9,46 @@ pub mod tx {
 
 pub mod types {
     pub use sui_sdk_types::*;
+    use {
+        crate::types::{parse_u64_value, strip_fields_owned},
+        serde::{de::Error as _, Deserialize, Serialize},
+    };
+
+    #[derive(Clone, Debug, PartialEq, Eq, Serialize)]
+    pub struct SuiBalance {
```
kouks: might be better to put this under `nexus_sdk::types` instead of here.

cli/src/tap/mod.rs
LKY-stephen: dag execute: default_agent execute
tap_execute: accessible agent execute
scheduler task create: default agent task
tap schedule-task: accessible agent-task

The prior two should be retired in the future, so we use tasks for everything.

sdk/src/events/parsing.rs
LKY-stephen: It depends on how we check it; the raw bytes one is really not for humans to read. And this one can ensure no fields are skipped? We can update it later to check if the skip or default value is respected.

cli/src/tap/mod.rs
kouks: Yeah this should definitely be improved in the future. Could even collapse it to just 1 command and a flag to use default agent. I can see having this many commands to execute be bad UX.

sdk/src/events/parsing.rs
kouks: > And this one can ensure no fields are skipped

But the type we're serializing is the one in the SDK so we cannot even compile if fields are skipped. If onchain changes, these tests will still pass but e2e will break.

Maybe add a note to these tests for now that they could be improved.

cli/src/tap/mod.rs
```diff
@@ -13,8 +13,8 @@ mod tap_register_skill;
 mod tap_registry;
 mod tap_requirements;
 mod tap_scaffold;
-mod tap_schedule;
-mod tap_schedule_address_funded;
+mod tap_schedule_task;
+mod tap_scheduled_task;
```
davidrotari19, line 17: What is the difference between them?

LKY-stephen: schedue task is to create a task, scheduled task is for manage/settle existing scheduled task. At least it is the goal. not complete yet.

cli/src/tap/mod.rs
LKY-stephen: Yes, that requires updating the main command layer. For now, I just ensure they are clear, we can redesign it in the future.

cli/src/tap/tap_scaffold.rs
LKY-stephen: added

cli/src/tap/tap_scheduled_task.rs
LKY-stephen: removed, I think it was due to scheduler/tap command layer diff. we should merge in near future.

sdk/src/events/mod.rs
LKY-stephen: Moved to serde.

sdk/src/nexus/crawler.rs
LKY-stephen: Correct, it is duplicated, removed

sdk/src/onchain_schema_gen/types.rs
LKY-stephen: added

sdk/src/transactions/dag.rs
LKY-stephen: no, paste error I think

sdk/src/transactions/dag.rs
LKY-stephen: removed

sdk/src/sui.rs
LKY-stephen: Moved

kouks (approved):

davidrotari19 (approved):

sdk/src/events/parsing.rs
LKY-stephen: Added a comment, the map between move struct and sdk types deserves a better check constraint.

PR #487, opened 2026-06-24
fix: make TAP demo flows work end-to-end

tuky191:
Six commits that get `just infra demo all` green against `nexus-next` main:

- `cli/tap`: treat the default DAG executor agent as shared; the agent is wrapped inside `DefaultDagExecutor` (dynamic field on the registry) so `sui_getObject(agent_id)` returns NotFound for it.
- `cli/tap schedule-task`: add `--dag-id` and route the default agent through `DefaultAddressFunded` so the Move-side `new_invoker_funded_default_agent_task` path is used.
- `cli/tap schedule-task`: only set `selected_dag` for runtime-selected skills; pinned skills must pass `None` per `agent_registry::resolve_agent_execution_config_dag`.
- `sdk/scheduler`: detect `task.owner == task.agent_id` in `enqueue_occurrence` and route through the new `add_occurrence_*_for_agent_funded_task` Move entries (added in nexus-next).
- `sdk/scheduler`: match the workflow-side `DagExecutionConfig` for the agent-bound execution policy lookup; the interface-side `AgentExecutionConfig` is destroyed before policy storage.
- `sdk/events/parsing`: allow `RequestScheduledOccurrenceEvent` from foreign emitters so PTBs that call into the scheduler from a user package (e.g. `demo_tap::trigger_delayed_occurrence`) aren't silently dropped.

## Test plan

- `cargo test -p nexus-sdk` and `cargo test -p nexus-cli` pass.
- `just infra demo all` against `nexus-next@fix/demo-default-dag-id` runs through `test_sum_default_agent.sh`, `test_sum_owned_agent.sh`, and `test_demo.sh` to completion.


cli/src/tap/tap_schedule_task.rs
```diff
@@ -120,20 +146,47 @@ pub(crate) async fn schedule_tap_task(
...
-            occurrence_budget,
-            selected_dag: Some(dag_id),
-            authorization_templates: Vec::new(),
-        },
+    // The default DAG executor agent is wrapped inside a dynamic field on the
+    // agent registry, so it can't be fetched as a top-level Sui object. The
+    // scheduler exposes `new_default_agent_task` which resolves the agent
+    // through the registry; pass `agent_id`/`skill_id` as None so the SDK
+    // builds that PTB instead of the regular agent-bound one.
+    let (params_agent_id, params_skill_id, tap_payment) = if is_default_dag_executor {
```
LKY-stephen, line 154: Well, for the default agent, currently we use Nexus DAG/scheduler for execution, not use Nexus tap. You want to unify everything to Nexus Tap?

sdk/src/idents/scheduler.rs
```diff
@@ -12,6 +12,28 @@ pub const SCHEDULER_MODULE: sui::types::Identifier =
 
 include!(concat!(env!("OUT_DIR"), "/idents_scheduler.rs"));
 
+/// Hand-written identifiers for scheduler entry points that aren't yet in the
+/// committed IR. Regenerate `generated/ir/scheduler.json` after the matching
+/// Move package update lands and drop this block.
```
kouks, line 17: You can regenerate these from your `nexus-next` branch - `CLAUDE.md` has the instructions so you should be able to just ask. Just make sure `nexus-next` is checked out on the correct branch, it spins up localnet, publishes the packages and regenerates.

sdk/src/nexus/scheduler.rs
```diff
@@ -45,6 +45,42 @@ impl ScheduledAgentExecutionConfig {
...
+        )
+        .await?;
+        match target.data.skill.dag_binding {
+            crate::types::SkillDagBinding::Pinned { dag_id } => Ok(dag_id),
+            crate::types::SkillDagBinding::RuntimeSelected => bail!(
+                "scheduled agent execution config for runtime-selected skill {skill_id} \
+                 of agent {agent_id} is missing the selected dag id"
+            ),
+        }
+    }
```
kouks, line 83: Probably safer to return a custom error that has a `::Structural` enum member which you can then check in `nexus-next` instead of string comparison.

LKY-stephen:
I think this is no longer an issue? we can close?

PR #489, opened 2026-06-26
feat: update for commit-settle change

LKY-stephen:
## Notable changes

- See the changelog, can check following points 

1. Review event parsing trust boundaries in `sdk/src/events/mod.rs` and `sdk/src/events/parsing.rs`: generated `EventWrapper<T>`, `DistributedEventWrapper<T>`, `RequestScheduledExecution<T>`, scheduled occurrence, request-walk unwrapping, distribution metadata preservation, and rejection of same-named foreign payloads.
2. Check `sdk/build.rs` generated-type adaptation rules: no broad replacement that hides real Sui framework types; serde attributes are narrow and tied to actual Move JSON shapes; generated `sui_framework` support is complete enough for IDs, UID, VecMap, table/queue/linked structures, and balance-bearing objects.
3. Validate generated support serde: human-readable address shapes and BCS for object IDs, `MoveOption`, `MoveString`, byte vectors, `TypeName`, `PolicySymbol`, scheduler numeric strings, TAP byte strings, and generated `NexusData` encoded fields.
4. Confirm stale manual public mirrors are gone and not rebuilt under another name: TAP/interface/scheduler/payment/event/data/port types should use generated aliases or helper impls on generated types, not separate manual enums/structs.
5. Review TAP/payment changes: `SkillPaymentPolicy` generated enum wrapper Move JSON, `ExecutionPayment` and `ScheduledPaymentReserve` deserialization, user-funded vs agent-funded max-budget behavior, settlement/refill/abort helpers, and CLI output compatibility.
6. Review scheduler APIs and tests: active-start selection for queue and periodic generator states, same shared-state type with different generator symbol, generated `PolicySymbol`, generated `TaskState`, scheduled request numeric fields, and foreign-emitter occurrence handling.
7. Review workflow/transaction builders for commit-then-settle and split-module adoption: function idents, object/pure argument API changes from `sui-transaction-builder` 0.3.1, `try_build`, type args, worksheet stamping, current committed-result settlement, record-only leader gas charge, expired abort, and refill flows.
8. Review CLI compatibility: command JSON should remain stable where documented, errors should be local before RPC for immutable/mutable agent misuse, and CLI TAP/scheduler commands should not expose retired endpoint-revision/config-digest/refund-mode fields.
9. Confirm Nexus leader can consume this SDK without local patches or local BCS parsers; if the Nexus repo still needs inline patches to use local SDK, call out whether that is intentional release wiring or a PR blocker.

## References

- [provide a list of tickets this PR resolves or references](https://github.com/Talus-Network/nexus/issues/818)

## Checklist

- [x] keep a changelog
- [x] write tests


cli/src/tap/mod.rs
```diff
@@ -591,6 +601,62 @@ pub(crate) enum PaymentsCommand {
...
+            value_name = "OBJECT_ID"
+        )]
+        agent_id: Option<sui::types::Address>,
+        #[command(flatten)]
+        gas: GasArgs,
+    },
+}
+
+#[derive(Subcommand)]
+pub(crate) enum SettleCommand {
```
kouks: should probably be called `ExecutionCommand`

LKY-stephen: The main reason is that there exists an execute command already. we may redesign subcommands after https://github.com/Talus-Network/nexus/issues/784?

cli/src/tap/tap_output.rs
```diff
@@ -168,10 +168,20 @@ pub(crate) fn requirements_result_json(result: &GetSkillRequirementsResult) -> s
...
 
+fn requirements_json(requirements: &nexus_sdk::types::SkillRequirements) -> serde_json::Value {
+    let mut value = serde_json::to_value(requirements).unwrap_or_else(|_| json!({}));
+    if let Some(object) = value.as_object_mut() {
+        if let Some(input_commitment) = object.remove("input_commitment") {
+            object.insert("input_schema_commitment".to_string(), input_commitment);
+        }
+    }
+    value
+}
```
kouks: why do we need to change this?

cli/src/tap/tap_payments.rs
```diff
@@ -178,6 +201,68 @@ async fn resolve_payment(
...
+
+    notify_success!(
+        "Refill transaction submitted: {digest}",
+        digest = result.tx_digest.to_string().truecolor(100, 100, 100),
+    );
+
+    json_output(&payment_refill_result_json(&result))
+}
+
+fn payment_refill_result_json(result: &RefillExecutionPaymentResult) -> serde_json::Value {
```
kouks: these functions usually go to `tap_output.rs`

cli/src/tap/tap_settle.rs
```diff
@@ -0,0 +1,171 @@
...
+
+fn execution_abort_result_json(result: &AbortExecutionResult) -> serde_json::Value {
+    json!({
+        "function": "abort_expired_execution",
+        "digest": result.tx_digest,
+        "tx_checkpoint": result.tx_checkpoint,
+        "dag_id": result.dag_id,
+        "execution_id": result.dag_execution_id,
+    })
+}
```
kouks, line 89: same for these with `tap_output.rs`

sdk/bin/regenerate_bindings.sh
```diff
@@ -0,0 +1,401 @@
+#!/usr/bin/env bash
```
kouks, line 1: can you add instructions to `CLAUDE.md` on how to use this?

sdk/src/events/parsing.rs
```diff
@@ -525,1760 +533,1129 @@ fn is_event_wrapper(tag: &sui::types::StructTag, objects: &NexusObjects) -> bool
...
-    impl<T> From<Option<T>> for MoveOptionBcs<T> {
-        fn from(value: Option<T>) -> Self {
-            Self {
-                vec: value.into_iter().collect(),
-            }
+    fn addr(byte: u8) -> sui::types::Address {
+        sui::types::Address::from([byte; 32])
+    }
+
+    fn iface_version(inner: u64) -> version::InterfaceVersion {
```
kouks: I'd stick with just `interface_version`

sdk/src/events/polling.rs
```diff
@@ -737,98 +737,3 @@ impl EventPoller {
         Ok(())
     }
 }
-
-#[cfg(test)]
```
kouks, line 830: why did we remove these tests?

kouks:
I'm not entirely happy with combining the types refactor and the logic for commit and settle in a single PR. I also think the move-binding integration for type parsing could be done much cleaner if a PR were to focus solely on that. Specifically I don't like 2 things:
1. The string replacements in the generated types in `build.rs`
2. The confusion created by prefixing/suffixing many helper functions with `_generated` - I think if we were to integrate move-binding types fully, we would only have generated types and this would just be naming clutter.

docs/cli.md
```diff
@@ -679,6 +679,42 @@ JSON output includes a `function` marker (`accomplish_tap_execution_payment` or
...
+**`nexus tap execution abort --execution-id <OBJECT_ID>`**
+
+Permissionlessly aborts an expired TAP DAG execution through the current workflow settlement abort entry. This is distinct from `nexus dag abort-expired-execution`, which still discovers and submits a ToolGas-assisted abort candidate.
+
+JSON output includes the `function`, `dag_id`, `execution_id`, and transaction `digest`/`tx_checkpoint`.
+
+{% hint style="info" %}
+This command requires that a wallet is connected to the CLI and holds sufficient SUI for gas.
+{% endhint %}
+
```
kouks, line 729: think we're missing docs for `tap schedule-task --dag-id`

sdk/src/nexus/workflow.rs
```diff
@@ -1212,6 +1621,36 @@ impl WorkflowActions {
             selected_candidate,
         })
     }
+
+    async fn submit_workflow_transaction(
```
kouks: this is a dupeof `submit_tap_transaction` in tap.rs - you can probably extract it to `NexusClient::submit_transaction`

cli/src/tap/mod.rs
kouks: The struct name is only internal no? And I think it makes sense that `tap execute` is to start an execution and `tap execution x` is to manage it somehow.

LKY-stephen: renamed

cli/src/tap/tap_output.rs
LKY-stephen: removed, some stale naming issue

cli/src/tap/tap_payments.rs
LKY-stephen: done, thx

cli/src/tap/tap_settle.rs
LKY-stephen: done, thx

sdk/src/events/parsing.rs
LKY-stephen: renamed

sdk/src/events/polling.rs
LKY-stephen: recovered

docs/cli.md
LKY-stephen: Added, it is basically just the old one, and with tap command, the dag-id is just an optional input

sdk/src/nexus/workflow.rs
LKY-stephen: done

sdk/bin/regenerate_bindings.sh
LKY-stephen: updated

LKY-stephen:
> I'm not entirely happy with combining the types refactor and the logic for commit and settle in a single PR. I also think the move-binding integration for type parsing could be done much cleaner if a PR were to focus solely on that. Specifically I don't like 2 things:
> 
> 1. The string replacements in the generated types in `build.rs`
> 2. The confusion created by prefixing/suffixing many helper functions with `_generated` - I think if we were to integrate move-binding types fully, we would only have generated types and this would just be naming clutter.

I agree, but the first point is a compromise due to serialization mismatch for move and Rust. It needs to fix move-binding, so before that, using this replace can at least centralize the change, and we know what we can delete in the future once move binding is updated. Regarding the second one, it is mostly some old type conflicts; I cleaned another round, should be better now. However, if we have time, I think we should update the module definition in types just like what we did with ident.

kouks (approved):

davidrotari19 (approved):

PR #513, opened 2026-07-17
feat/update-for-new-verifier

LKY-stephen:
## Notable changes

- see CHANGELOG.md

## Checklist

- [X] keep a changelog
- [X] write tests


sdk/src/move_bindings/extensions/network_auth.rs
```diff
@@ -31,12 +30,10 @@ impl IdentityKey {
         }
     }
 
-    /// Construct a generated Move `IdentityKey::Tool` value from a tool FQN string.
-    pub fn tool_fqn(fqn: &str) -> Self {
+    /// Construct a generated Move `IdentityKey::Tool` value from its stable Tool object ID.
+    pub fn tool(tool_id: sui::types::Address) -> Self {
```
davidrotari19, line 34: Tool_id is a derived object of `FQN `and `Registry`, is there a benefit of handling `tool_id` directly?

sdk/src/move_boundary/mod.rs
```diff
@@ -110,11 +110,34 @@ impl<'a> NexusPtbBuilder<'a> {
         )?)
     }
 
+    /// Add a runtime-owned Move call with already validated type arguments.
```
davidrotari19, line 113: Is this needed?

sdk/src/move_boundary/mod.rs
```diff
@@ -110,11 +110,34 @@ impl<'a> NexusPtbBuilder<'a> {
...
     /// Build a Move `0x1::ascii::String` from bytes.
     pub fn ascii_string(&mut self, value: impl AsRef<str>) -> Result<Argument, BuildError> {
         ascii_string(&mut self.tx, value)
     }
 
+    /// Build a Move `0x1::string::String` from UTF-8 bytes.
+    pub fn move_string(&mut self, value: impl AsRef<str>) -> anyhow::Result<Argument> {
+        let bytes = self.tx.arg(&value.as_ref().as_bytes().to_vec())?;
+        self.call_target(
+            || CallTarget::new(move_std::call_package(), "string", "utf8"),
```
davidrotari19: I think move-binding already gives you the function

sdk/src/move_bindings/extensions/network_auth.rs
LKY-stephen: Tool_id is better for code operations, and its uniqueness is kind of automatically granted compared to a single string. I think using a FQN name is not for code, but only for display; that's why I lean to use id when we are handling resources.

sdk/src/move_boundary/mod.rs
LKY-stephen: This is for registering an external verifier, which needs a third-party verifier only known at runtime (in strings). Also used for   verifier_registry::add_obj<ObjectType>

sdk/src/nexus/registered_key.rs
```diff
@@ -1,453 +1,165 @@
...
-    if output_variant == ERR_EVAL_VARIANT {
-        return Ok(hash_bytes(&terminal_err_eval_reason_bytes(
-            output_variant,
-            output_ports_data,
-        )?));
-    }
-
-    canonical_output_payload_sha256(output_variant, output_ports_data)
+/// Tool signature message: `leader_signature || SHA-256(result_bytes)`.
+pub fn tool_signature_message(
```
davidrotari19, line 83: Leader signature is potentially redundant for onchain authorization, it seems more like on offchain property

sdk/src/signed_http/mod.rs
```diff
@@ -1,39 +1,15 @@
...
-//!
-//! # Versions
-//! - [`v1`] is the currently implemented protocol version.
-
-pub mod v1;
+//! Minimal Leader-to-Tool signing transport.
+//!
+//! The protocol deliberately signs only the values consumed by the onchain RegisteredKey
+//! verifier:
+//! - the Leader signs `SHA-256(BCS(canonical_tool_inputs))`;
```
davidrotari19, line 5: In the minimal form, this seems redundant. The present leader signature is justified only by offchain caller authentication. That seems not to be required given what is stated below, the minimal (tool_id, input_hash, output_hash) commitment and let leader_cap handle leader authorization.

sdk/src/signed_http/mod.rs
```diff
@@ -1,39 +1,15 @@
...
+//! - the Tool signs `leader_signature || SHA-256(result_bytes)`;
```
davidrotari19: This seems to only say, this leader key signed these 32 bytes, and this tool key signed that leader signature plus this output hash. This does not seem like Tool T, executing operation X in execution E, received input I and returned output O. The new protocol is effectively a simple chained message signature whose meaning comes primarily from ambient onchain context. For the same keys, input hash and output this can be replayed. Is this the goal?

sdk/src/signed_http/mod.rs
```diff
@@ -1,39 +1,15 @@
...
-//! - no bespoke body canonicalization is required,
-//! - the signature is stable and auditable (store claims bytes + signature).
-//!
-//! # Versions
-//! - [`v1`] is the currently implemented protocol version.
-
-pub mod v1;
+//! Minimal Leader-to-Tool signing transport.
+//!
+//! The protocol deliberately signs only the values consumed by the onchain RegisteredKey
```
davidrotari19: This argument seems circular, the verifier consumes only those fields because we designed the protocol that way

sdk/src/nexus/registered_key.rs
LKY-stephen: leader sig is used to preserve the binding (leader sig, input, output, tool), without it, the result from an old query or other leaders.

sdk/src/signed_http/mod.rs
```diff
@@ -1,39 +1,15 @@
...
+//! verifier:
+//! - the Leader signs `SHA-256(BCS(canonical_tool_inputs))`;
+//! - the Tool signs `leader_signature || SHA-256(result_bytes)`;
+//! - `result_bytes` is the exact BCS `TaggedOutput` submitted onchain.
```
davidrotari19, line 7: I think we have two options:
- The tool certifies: `For canonical input I, I attest to output O`. This is reusable as its not bounded. Then we dont need leader signature, leader_cap authorizes who may submit it. 
- The tool certifies: For this particular workflow invocation, requested by this leader, I returned this result.

What we have now seems to be in between these two

sdk/src/signed_http/mod.rs
LKY-stephen: The goal is onchain call wants to verify the leder submit the result from the certsin tool goven corresponding inputs. So thr tool technically need to sign the three things input, leader, output (they don't have context for anything else, otherwise they need to sign more). So i used this architecture. Technically tool can also sign leader cap id but it is also not in the contrxt of tool, that's why i added the leader sig which is needed anyway offchain and can bind the leader key with leader id naturally. I guess we can have other form, if you have a better shape we can change

sdk/src/signed_http/mod.rs
LKY-stephen: I think we are the second?

sdk/src/signed_http/mod.rs
davidrotari19: I think we’re mixing two properties. `leader_cap` and the Sui transaction already prove the authorized onchain submitter, while the Tool key proves the Tool. To bind input and output, the Tool can sign domain || input_hash || output_hash. Including leader_signature additionally requires the HTTP caller and onchain submitter to be the same leader, which may not
hold because invocation and submission are separate operations(what happens when offchain tool is costly operation, the first leader send the request but failed to send the transaction, should we redo the costly compuation?). It also does not prevent old query reuse: the same keys(which are static), input(the input can be the same for two workflows), and output(can be the same given same input) produce the same proof.

sdk/src/nexus/registered_key.rs
davidrotari19: >old query

But it does not prevent old query reuse: the same keys, input,and output produce the same proof. Those easy can be the same

>other leaders

I was thinking the following. We have an offchain tool that does an expensive operation, first leader sends the request, gets the response but fails before sending it onchain. What does second leader needs to do? should redo the request(it replicates the same expensive operation), does it mean offchain tool needs to charge double(it did the expensive operation twice) because leader mistake

sdk/src/signed_http/mod.rs
davidrotari19: Our canonical input is:` I = SHA256(BCS(vector<CanonicalToolInput>))` where each item is { port_name: bytes, data: NexusData { storage, one, many } }, sorted by port name bytes. It contains the effective vertex inputs, defaults plus upstream outputsmbut not tool_id, execution_id, walk_index, vertex, leader, nonce, or expiry. It is also not the JSON body the Tool executes.Therefore, the same port values produce the same `I` across workflow invocations. So we are not fully in the second model: the proof is leader bound, but not invocation bound.

sdk/src/signed_http/mod.rs
LKY-stephen: If the first leader failed and second leader continue, the tool can reuse the output but sign the new envelope to highlight the second leader also called the tool.

i add the leader basically to prevent the replay of proof. What nexus wants to confirm is the four elements binding. Reuse proof is the wrong thing. Signature never prove the computation, it only prove the invocation, so sender is a part of the statement.

In future, if two tool invocations have the same tool signature, we know it is a replay and the query is not really executed.

davidrotari19: A signature only proves that a key endorsed some bytes, it does not inherently prove a new invocation or computation. Binding (leader, tool, input, output) is not enough to identify an invocation: two genuine calls by the same leader with the same input/output produce the same deterministic signature.

sdk/src/signed_http/mod.rs
LKY-stephen: Yes, not fully. I only pick what onchain info required and tool can have access: pubkeys, input, output. The onlything i think we can add is the onchain timestamp but that related to the clock asynchrony so i feel better not to add.

LKY-stephen: Let me know if you think any additional onchain info can be certified semantically offchain, we can add to aux

sdk/src/signed_http/mod.rs
LKY-stephen: correct, it is not enough, and it is the only thing offchain tool can prove to my understanding? Maybe we can add a timestamp, but that is not reliable, and other info cannot be verified onchain. Fundamentally, it is a kind of Man in the middle case. I thought this is the best effort the tool can prove through sig.

sdk/src/signed_http/mod.rs
davidrotari19: We can do better without introducing clock synchronization. The onchain context already known when invoking the Tool includes tool_id, execution_id, walk_index, and runtime_vertex. We could derive or store a unique invocation_id from this contextand have the Tool sign domain || invocation_id || input_hash || output_hash. The contract then verifies the invocation ID against the execution(we can derive this fully oncahain).  These fields must be included in the Tool signed message and checked onchain, not merely added to aux.

sdk/src/signed_http/mod.rs
LKY-stephen: But if we assume all tools have clock with bounded delay, then we can add a timestamp

sdk/src/signed_http/mod.rs
LKY-stephen: But how can the tool verify these info? We don't assume tool can read sui?  If so we are basically ask the leader to sign a random field in addition to the payload

davidrotari19: The Tool does not need to verify that invocation_id is correct onchain. Tool cannot verify that the supplied input_hash corresponds to the Sui execution, so tool already signs to a random input(leader also signs a random one).  It only needs to bind its response to the exact context presented in the authenticated request. The Tool does not need Sui access. If the leader supplies a random or incorrect ID, the resulting proof is simply rejected onchain.

sdk/src/move_boundary/mod.rs
LKY-stephen: removed

sdk/src/nexus/registered_key.rs
LKY-stephen: updated with the discussed signature

sdk/src/signed_http/mod.rs
LKY-stephen: updated with the discussed signature

sdk/src/signed_http/mod.rs
LKY-stephen: updated with the discussed signature

sdk/src/signed_http/mod.rs
LKY-stephen: reword a bit

sdk/src/signed_http/mod.rs
LKY-stephen: updated with the discussed signature

sdk/src/signed_http/mod.rs
davidrotari19: The protocol should be derived from the guarantees we want, not derive the guarantees from the protocol. Given the new verifier this is irrelevant

davidrotari19 (approved):

PR #540, opened 2026-08-10
[Bug]Protocol event handling

davidrotari19:
## Notable changes

- see `CHANGELOG`

## References

- solves https://github.com/Talus-Network/nexus/issues/887

## Checklist

- [x] keep a changelog
- [x] write tests
- [x] create tickets for `TODO: <>` statements


sdk/src/events/parsing.rs
```diff
@@ -44,11 +46,21 @@ pub(super) fn decode_nexus_event(
...
     wrapper_type: &sui::types::StructTag,
     objects: &NexusObjects,
-) -> anyhow::Result<Option<NexusEvent>> {
+) -> Result<Option<NexusEvent>, NexusEventDecodeError> {
     let Some(event_type) = NexusEventType::resolve(wrapper_type, objects) else {
         return Ok(None);
     };
-    let (data, distribution) = parse_bcs(event_type.name, contents)?;
+    if !supports_event(event_type.name) {
+        if objects.is_active_emitter(emitting_package) {
```
LKY-stephen: I remember we allowed an external package to emit the package? So this check should be for the definition package?

sdk/src/transactions/tool.rs
```diff
@@ -644,6 +644,34 @@ pub fn update_tool_timeout_ptb(
...
+/// Builds a transaction that replaces an off chain Tool URL.
+///
+/// The `tool` is the current shared Tool reference and `owner_cap` must grant
+/// authority over that Tool.
+///
+/// # Errors
+///
+/// Returns an error when an object argument, URL argument, or generated Move
+/// call cannot be added to the transaction.
+pub fn update_off_chain_tool_url_ptb(
```
LKY-stephen, line 656: do we have cli for this?

davidrotari19: No, this operation should not be allowed in direct sense. Changing the url of a tool, should not be a safe operation to do. This needed change in a new pr

LKY-stephen: I remember we have a leader mechanism to check the url and active the tool again after the URL changes? So a URL change will first make the tool inactive

sdk/src/events/parsing.rs
davidrotari19: `resolve()` already validates the inner event’s defining package. Do you mean that once `resolve()` succeeds, unsupported events should always error regardless of `emitting_package`?

sdk/src/transactions/tool.rs
davidrotari19: I meant, changing the tool url in place would imply that URL is mutable deployment, which in my opinion should not be allowed. We need a proper mechanism how to handle changes in the tool, current update in place is way too lose to allow it, and `verified` does not make tool inactive, its a flag that is ignored by leader and workflow

sdk/src/events/parsing.rs
LKY-stephen: I mean the code seems checking if this is an event from the old nexus packages, so we should check the "definition_package" not emitting_package?

davidrotari19: [f78d828](https://github.com/Talus-Network/nexus-sdk/pull/540/commits/f78d828196e1b41f739a6932b46d1e0cdae94765)

LKY-stephen (approved):
LGTM

PR #549, opened 2026-08-12
use receiving mode in priority fee vault

LKY-stephen:
## Notable changes

- see CHANGELOG.md
## References

- https://github.com/Talus-Network/nexus/issues/896

## Checklist

- [X] keep a changelog
- [X] write tests


cli/src/network/actions.rs
```diff
@@ -1,4 +1,57 @@
-use crate::{command_title, display::json_output, loading, notify_success, prelude::*, sui::*};
+use {
+    crate::{command_title, display::json_output, loading, notify_success, prelude::*, sui::*},
+    nexus_sdk::nexus::network::CollectPriorityFeeDepositsResult,
+};
+
+pub(crate) async fn collect_priority_fees(
```
davidrotari19, line 6: why we need leader_cap_id, do we want to do a subset and only for a `leader`?

sdk/src/nexus/network.rs
```diff
@@ -32,6 +47,25 @@ pub struct WithdrawPriorityFeeResult {
     pub tx_digest: sui::types::Digest,
 }
 
+/// A deposit left under the vault because its embedded leader is no longer registered.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct SkippedPriorityFeeDeposit {
```
davidrotari19, line 52: what happens with these funds?

sdk/src/nexus/network.rs
```diff
@@ -225,6 +259,245 @@ impl NetworkActions {
...
+        .get_object_owned_objects::<PriorityFeeDepositV2>(vault_id, object_type)
+        .await
+        .map_err(|error| {
+            NexusError::Rpc(anyhow::anyhow!(
+                "Failed to discover priority fee deposits owned by vault '{vault_id}': {error}"
+            ))
+        })?;
+
+    for deposit in &deposits {
+        if deposit.data.id.id.bytes != deposit.object_id {
```
davidrotari19: why this failure?

sdk/src/nexus/network.rs
LKY-stephen: Locked for now because they do not have leader_cap to claim; we may decide what to do in the future.

cli/src/network/actions.rs
LKY-stephen: This is for helping a leader to collect his own deposit and then he can swap and withdarwal at once (also not pay gas for others' priority fee). There is a collect_priority_fee_deposit_batch which can be used for general collect

sdk/src/nexus/network.rs
LKY-stephen: removed

davidrotari19 (approved):

PR #574, opened 2026-09-02
[TAP]run TAP tests with published Nexus bytecode

davidrotari19:
## Notable changes

- see `CHANGELOG`

## Checklist

- [x] keep a changelog
- [x] write tests
- [x] create tickets for `TODO: <>` statements

### DIFF
--- a/cli/src/tap/mod.rs
+++ b/cli/src/tap/mod.rs
@@ -13,6 +13,9 @@ mod tap_registry;
 mod tap_requirements;
 mod tap_scaffold;
 mod tap_settle;
+mod tap_test;
+mod tap_test_overlay;
+mod tap_test_vm;
 mod tap_update_skill;
 mod tap_validate_skill;
 mod tap_vault;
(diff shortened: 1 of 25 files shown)


cli/src/tap/templates/skill_config.json.jinja
```diff
@@ -7,6 +7,5 @@
     "schedule_policy": "Once",
     "fixed_tools": []
   },
-  "shared_objects": [],
```
LKY-stephen, line 10: So now we need to use entry_group/tool_output for deciding share objects to use for onchain tools?

cli/src/tap/mod.rs
```diff
@@ -104,6 +108,42 @@ use {
...
+            value_parser = ValueParser::from(expand_tilde),
+            default_value = "."
+        )]
+        path: PathBuf,
+        #[arg(
+            long = "build-env",
+            short = 'e',
+            value_enum,
+            default_value_t = TapTestEnvironment::Testnet,
+            help = "Published Nexus environment used by the test VM.",
```
LKY-stephen, line 130: May need values examples for input? What should the user provide: "testnet, mainnet"?

LKY-stephen (approved):

cli/src/tap/templates/skill_config.json.jinja
davidrotari19: This was never a field. I removed it as SkillConfig did not have a shared_objects, so this was ignored. Everything flows through the DAG

cli/src/tap/mod.rs
davidrotari19: It already prints it, if you do `nexus tap test --help`

